PR #1, opened 2018-09-27
Add skeleton config for flutter

hinoka:
This is used for LUCI


hinoka:
@cbracken @gspencergoog @shishkander ptal

devoncarew:
As a drive by comment, @athomas and @whesse from the Dart infra team are likely familiar with these types of files, and could be good reviewers for future PRs.

PR #2, opened 2019-01-12
fix triggers, give framework short name

dnfield:
Right now the engine is getting triggered by flutter/flutter instead of flutter/engine

Also adds a short name for the framework.

I'm not 100% sure this will work - partly an experiment here.


cbracken:
lgtm

PR #8, opened 2019-02-05
caches

dnfield:
We need this for LUCI to recognize these directories in the recipe are actually cache dirs that are meant to be reused on the bots.


riannucci:
lgtm!

PR #12, opened 2019-02-07
Use Ubuntu-16.04

dnfield:
I strongly suspect that our Java certificate issues will go away if we bump Ubuntu.  Will revert if this breaks anything

@riannucci


tvolkert:
LGTM

PR #16, opened 2019-02-12
Use Ubuntu-16.04

dnfield:
Once internal support lands, we'll be able to use Ubuntu 16, which is more consistent with other parts of our Cirrus/Docker infra.

__this change should not land until the CL for upgrading us to Ubuntu machines does, but once that CL lands we will be required to use this one!__


dnfield:
Bots are coming online https://chromium-swarm.appspot.com/botlist?c=id&c=task&c=os&c=status&d=asc&f=pool%3Aluci.flutter.prod&k=pool&s=id

PR #22, opened 2019-03-21
Shard engine

dnfield:
This will enable sharding once https://chromium-review.googlesource.com/c/chromium/tools/build/+/1534736 lands.


dnfield:
@riannucci - is there a better way to manage this config?

dnfield:
(Per some other LUCI team members, there's nothing today for it but there's something coming down the pipe at some point.  for now this is what it is)

PR #32, opened 2019-04-12
Make packaging build only the latest commit

dnfield:
Also puts Windows bots to 6 concurrent invocations like everyone else.


Hixie:
RSLGTM

dnfield:
@digiter FYI

PR #33, opened 2019-05-15
Notify blamelist on new failures

dnfield:
Just got a notification from LUCI that I broke another team's build,and it was actually helpful to be so quickly notified.

This will send multiple notification emails if someone breaks multiple builds (e.g. a bad engine patch that breaks 10 engine builders will send 10 emails).


digiter:
LGTM

nit: It's not clear from the config, who will receive the notification emails?

dnfield:
Email will go out to the blamelist for the commit.  If you click the blamelist tab it's that list:

image

PR #37, opened 2019-06-19
Remove configurations for dynamic patching

jason-simmons:


dnfield:
/cc @digiter  FYI.

It would be nice to split some more of the existing ones up further - we still have a couple jobs that take 30-35 minutes,  that could be more like 15.

digiter:
If I understand correctly, this is related to https://github.com/flutter/flutter/issues/31455.

dnfield:
Yes.

PR #38, opened 2019-06-25
add LUCI triggers, jobs, and builders for building ios usb dependencies

christopherfujino:


config/cr-buildbucket.cfg
```diff
@@ -173,6 +173,72 @@ buckets {
         properties_j: "$depot_tools/osx_sdk:{\"sdk_version\": \"10e125\"}" # Xcode 10.2
       }
     }
+    builders {
+      name: "ideviceinstaller"
+      mixins: "mac"
+      recipe {
+        name: "flutter/ios-usb-dependencies"
+        properties: "shard:tests"
```
dnfield: What does this do? Are there other values we'd want to set?

christopherfujino: Actually, that's a property i copied from the flutter/flutter recipe which i'm not using in my recipe. I'll delete them.

christopherfujino: Done

digiter:
And https://github.git.corp.google.com/ actually already contains the mirrors for these repos, e.g.  https://github.git.corp.google.com/libimobiledevice/ideviceinstaller/.

christopherfujino:
@digiter 
> And https://github.git.corp.google.com/ actually already contains the mirrors for these repos, e.g. https://github.git.corp.google.com/libimobiledevice/ideviceinstaller/.

True, but homebrew (which the recipe uses to build the packages) doesn't support authenticated git repos, and also we wouldn't be able to maintain our patches on those mirrors.

Also, I wasn't planning to surface these repos to the end users, but because of [this issue](https://github.com/flutter/flutter/issues/35102), users are now installing these dependencies from our mirrors using homebrew.

PR #43, opened 2019-07-30
Use lucicfg to generate our config

dnfield:
This will make it easier to manage our builders when we ahve to add or remove them - strings are concentrated in one place now.

This does make a few minor modifications to our current config:

- Fixes a typo in `cocapods` for a cache
- Fixes a typo in a repo url for ios-deploy
- Explicitly specifies some properties that were relying on default values

Otherwise, it just reorders a few things.

Fixes https://github.com/flutter/flutter/issues/37190


README.md
```diff
@@ -7,4 +7,4 @@ Contains configuration and tools related to Flutter's build infrastructure.
 One may speed up the propagation of these configuration files by visiting the
 [luci-config web UI](https://luci-config.appspot.com/), logging in, and
 searching for projects/flutter. From there, one may click on the
-projects/flutter search result and click the download icon to force an update.
\ No newline at end of file
+projects/flutter search result and click the download icon to force an update.
```
christopherfujino: this is a weird diff?

config/main.star
```diff
@@ -0,0 +1,327 @@
...
+  if cores != None:
+    dimensions['cores'] = cores
+
+  luci.builder(
+    name = name,
+    bucket = 'prod',
+    executable = recipe,
+    properties = properties,
+    service_account = 'flutter-prod-builder@chops-service-accounts.iam.gserviceaccount.com',
+    execution_timeout = 3 * time.hour,
```
christopherfujino: is this a new timeout, or was it the default already?

dnfield: Existing delay

README.md
dnfield: I have no idea haha.  I should probably add something here about main.star though.

config/main.star
```diff
@@ -0,0 +1,327 @@
...
+    {
+      '$depot_tools/osx_sdk': {
+        'sdk_version': '10e125', # 10.2
+      },
+    },
+    properties
+  )
+  if caches == None:
+    caches = []
+  caches.append(swarming.cache('osx_sdk'))
```
athomas: This modifies the input list, probably better to create a copy, because if you do something like this later the results might be unexpected:
MAC_CACHES = [...]
mac_builder(caches=MAC_CACHES, ...)
mac_builder(caches=MAC_CACHES, ...) # <- this builder will have two osx_sdk caches in the list.

config/cr-buildbucket.cfg
```diff
@@ -1,338 +1,619 @@
...
-    identity: "luci-scheduler@appspot.gserviceaccount.com"
-  }
-  acls {
-    role: SCHEDULER
-    group: "project-flutter-prod-schedulers"
-  }
-  acls {
-    # Allow luci-migration to bump next build number.
-    role: WRITER
-    identity: "luci-migration@appspot.gserviceaccount.com"
```
athomas: Is the removal of everything related to luci-migration intentional? (I would assume so, but just checking).

config/main.star
```diff
@@ -0,0 +1,327 @@
...
+COMMON_FRAMEWORK_BUILDER_ARGS = {
+  'recipe': 'flutter/flutter',
+  'console_view_name': 'framework',
+  'triggered_by': ['master-gitiles-trigger-framework'],
+  'triggering_policy': scheduler.greedy_batching(max_concurrent_invocations=6)
+}
+
+linux_builder(name='Linux|frwk', properties={'shard': 'tests'}, **COMMON_FRAMEWORK_BUILDER_ARGS)
+linux_builder(name='Linux Coverage|lcov', properties={'shard': 'coverage', 'coveralls_lcov_version': '5.1.0',}, **COMMON_FRAMEWORK_BUILDER_ARGS)
+mac_builder(name='Mac|frwk', properties={'shard': 'tests', 'cocoapods_version': '1.6.0'}, caches=[swarming.cache(name='flutter_cocoapods', path='cocoapods')], **COMMON_FRAMEWORK_BUILDER_ARGS)
```
athomas: The cache path changed to cocoapods (the old path was cocapods), looks like a fixed typo. Was this intentional?

config/main.star
```diff
@@ -0,0 +1,327 @@
...
+windows_builder(name='Windows|frwk', properties={'shard': 'tests'}, **COMMON_FRAMEWORK_BUILDER_ARGS)
+
+COMMON_ENGINE_BUILDER_ARGS = {
+  'recipe': 'flutter/engine',
+  'console_view_name': 'engine',
+  'triggered_by': ['master-gitiles-trigger-engine'],
+  'triggering_policy': scheduler.greedy_batching(max_batch_size=1, max_concurrent_invocations=3)
+}
+
+def engine_properties(build_host=False, build_android_debug=False, build_android_aot=False, build_android_vulkan=False, build_ios=False, needs_jazzy=False):
```
athomas: I can see more builders getting "build_<something>:false" properties than before because of these defaults, is that intentional?

e.g. Windows Host Engine now gets:
        properties_j: "build_android_debug:false"
        properties_j: "build_android_vulkan:false"
        properties_j: "build_ios:false"

config/main.star
```diff
@@ -0,0 +1,327 @@
...
+
+  luci.builder(
+    name = name,
+    bucket = 'prod',
+    executable = recipe,
+    properties = properties,
+    service_account = 'flutter-prod-builder@chops-service-accounts.iam.gserviceaccount.com',
+    execution_timeout = 3 * time.hour,
+    dimensions = dimensions,
+    swarming_tags = ['vpython:native-python-wrapper'],
```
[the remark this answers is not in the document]
dnfield: Thanks, will remove.

config/main.star
dnfield: Good catch, will change that.

config/cr-buildbucket.cfg
dnfield: Yes, some of it was making lucicfg complain anyway

config/main.star
dnfield: Yes, the old cache never worked and I finally know why!

config/main.star
dnfield: These were previously just implicit defaults.  Now they're explicit.

PR #48, opened 2019-08-09
shard ios further for bitcode

dnfield:
Bitcode builds will be much slower since we can't use goma.  This shards out debug/release/profile for changes in https://chromium-review.googlesource.com/c/chromium/tools/build/+/1746669 

We shouldn't land this until after that - it's safe to land beforethen, but would have three macs doing the same work until the sharding logic is present.


eseidelGoogle:
Curious: Does Chrome build for Bitcode?  If so, do they also not use GOMA?

dnfield:
I haven't asked anyone directly, but from my various searches on the topic it's not possible to run goma using Xcode, and you have to use Xcode to build a bitcode enabled binary.  Chrome may have some option to work with this, I'll try to ask more directly about it.

dnfield:
Chrome has options to build using bitcode, but it looks like they don't ship anything with bitcode. Nico Weber confirmed offline that they can't use GOMA for their bitcode builds, which is one reason they don't build for bitcode.  It's a bit different for us since we're vending a framework to be consumed downstream and linked in though...

PR #56, opened 2019-09-12
use all linux bots now

dnfield:
The Fuchsia (Debian) bots are now capable of building the engine


config/cr-buildbucket.cfg
```diff
@@ -92,7 +92,7 @@ buckets: <
       swarming_host: "chromium-swarm.appspot.com"
       dimensions: "cores:8"
       dimensions: "cpu:x86-64"
-      dimensions: "os:Ubuntu-16.04"
```
godofredoc: Should we also delete the ubuntu bots and keep only the debian ones? based on usage we can request more debian bots if needed.

dnfield: If we can replace them that's fine, but I'd rather not diminish capacity if we can avoid it right now.

PR #58, opened 2019-09-13
Add drone builders

dnfield:
I made a new recipe, want to have builders that are set to use it by default.  These are meant to run builds and then isolate build outputs for other builders to package.


digiter:
If I understand correctly, the purpose of this PR is to separate builds from other tasks, so that we can use powerful machines later for building, is that right?

dnfield:
It's so we can make a request to a machine that _only_ does a build (and then isolates certain files/directories for later usage).  Our current engine recipe is more complicated and does a ton of other stuff.

So at a high level, yes, more or less that's the case.

PR #59, opened 2019-09-16
set fuchsia_ctl version

dnfield:
This needs to land before https://chromium-review.googlesource.com/c/chromium/tools/build/+/1808258


dnfield:
I mistakenly published version:0.0.2 with binaries that were compiled on mac instead of linux :(  0.0.3 is good.

PR #60, opened 2019-09-20
bump fuchsia_ctl

dnfield:
I messed up tagging v0.0.4 - two packages now have that.  I'll try to get that removed from v0.0.3, but in the mean time I just added v0.0.4+1 to v0.0.4 in cipd.

This updates the package so it actually prints test output.  Test run can be seen here:  https://chromium-swarm.appspot.com/task?id=4764bf89bfe7a810

There's a bug in the recipe code where it isn't properly setting an exit code on failure.  I'll look into that separately.


iskakaushik:
LGTM

PR #63, opened 2019-10-01
Only enable jit_release on Android debug shard

jonahwilliams:
Since the value defaults to True, it was being built on all shards


jonahwilliams:
Do I need to do anything to publish this, or does it get picked up automagically?

dnfield:
Magically - follow along at https://luci-config.appspot.com/#/projects/flutter

The change hasn't replicated yet to GoB though

PR #64, opened 2019-10-02
bump fuchsia_ctl

dnfield:
This will pick up flutter/packages#41 - so tests will correctly set the exit code when they fail.


dnfield:
Going to land this TBR to get CI properly failing ASAP.

PR #66, opened 2019-10-09
Format main.star

digiter:


config/main.star
```diff
@@ -19,136 +21,156 @@ FLUTTER_GIT = 'https://chromium.googlesource.com/external/github.com/flutter/flu
...
-  config_dir = '.',
-  tracked_files = [
-    'cr-buildbucket.cfg',
-    'luci-logdog.cfg',
-    'luci-milo.cfg',
-    'luci-notify.cfg',
-    'luci-scheduler.cfg',
-    'project.cfg',
-  ],
+    config_dir='.',
```
dnfield: This looks like it got four spaces instead of two, and it's not entirely clear why.

config/main.star
```diff
@@ -19,136 +21,156 @@ FLUTTER_GIT = 'https://chromium.googlesource.com/external/github.com/flutter/flu
...
-luci.logdog(
-  gs_bucket = 'chromium-luci-logdog',
-)
+luci.logdog(gs_bucket='chromium-luci-logdog',)
 
 luci.milo(
-  logo = 'https://storage.googleapis.com/chrome-infra-public/logo/flutter-logo.svg',
-  favicon = 'https://storage.googleapis.com/flutter_infra/favicon.ico',
+    logo=
+    'https://storage.googleapis.com/chrome-infra-public/logo/flutter-logo.svg',
```
dnfield: These two lines are harder to read.

config/main.star
dnfield: Hm.  It looks like yapf wants to make this like python, where using double indendtation for line breaking reasons is common.  It looks strange to me but ... *shrug*

digiter:
Probably work on this later when I got time.

PR #71, opened 2019-11-20
[luci] Decommision the coverage shard for Flutter Framework.

digiter:
Per https://github.com/flutter/flutter/issues/31615, we'd like to run coverage along framework tests and remove this Linux coverage shard.


digiter:
Seems not from my understanding in the recipes.

dnfield:
So we won't be uploading coverage data to coveralls anymore?

digiter:
In my understanding, we don't upload coverage currently. [This line](https://chromium.googlesource.com/chromium/tools/build.git/+/master/scripts/slave/recipes/flutter/flutter.py#231) has no effect as there's no shard named "coverage".

dnfield:
Ahhh - that shard used to be named coverage I believe.  It was broken by https://github.com/flutter/infra/pull/69 - we didn't update it for the recipe there.

I'm not really clear on who was using coveralls anyway, and my impression was that it wasn't accurately reporting coverage.  We should probably have a bug open somewhere to actually check coverage of the framework though.

digiter:
SG, moved the discussion into https://github.com/flutter/flutter/issues/45271.

PR #73, opened 2019-12-11
Builder configs for cocoon try jobs.

godofredoc:
This is to enable presubmit tests for cocoon repo on luci.

https://github.com/flutter/flutter/issues/43348


config/main.star
```diff
@@ -15,6 +15,7 @@ The documentation for lucicfg can be found here:
 https://chromium.googlesource.com/infra/luci/luci-go/+/refs/heads/master/lucicfg/doc/README.md
 """
 
+COCOON_GIT = 'https://chromium.googlesource.com/external/github.com/flutter/cocoon'
 FLUTTER_GIT = 'https://chromium.googlesource.com/external/github.com/flutter/flutter'
 ENGINE_GIT = 'https://chromium.googlesource.com/external/github.com/flutter/engine'
```
digiter: No need to address in this PR, do we want to move the repos out of chromium?

config/main.star
```diff
@@ -298,6 +305,18 @@ linux_try_builder, linux_prod_builder = linux_builder()
...
 
+COMMON_COCOON_BUILDER_ARGS = {
+  'recipe': 'flutter/cocoon',
+  'console_view_name': 'cocoon',
+  'list_view_name': 'cocoon-try',
+}
+
+COCOON_EXTRAS = {
+  'caches': [swarming.cache(name='dart_pub_cache', path='.pub-cache')],
+}
```
digiter: Why this is in COCOON_EXTRAS, not COMMON_COCOON_BUILDER_ARGS?

config/main.star
godofredoc: yes, but will need a little bit of experimentation and probably a bunch of new gerrit configs.

config/main.star
godofredoc: Good point I was following the convention on the other configs but given that cocoon will only have linux configs it doesn't make sense to split the configs.

PR #74, opened 2020-01-06
Reland "Migrate to Xcode 11.0 so that the iOS 13 SDK may be used. (#62)"

LongCatIsLooong:
This reverts commit f53b002167c06751ef7a603d14f5a8d9e44a178c.

See: https://github.com/flutter/engine/pull/15154.


config/main.star
```diff
@@ -258,7 +258,7 @@ def mac_builder(properties = {}, caches=None, category = 'Mac', **kwargs):
   properties = merge_dicts(
     {
       '$depot_tools/osx_sdk': {
-        'sdk_version': '10e125', # 10.2
+        'sdk_version': '11a420a', # 11.0
```
dnfield: Is this right?  Are we on 11.0 or 11.1?

LongCatIsLooong: Per https://en.wikipedia.org/wiki/Xcode it's 11.0. I'll change the title.

LongCatIsLooong:
Whoops I accidentally pressed enter when I was trying to change the commit message. @cbracken should I revert and change the commit message?

cbracken:
I wouldn't bother :) It's just added convenience, not the end of the world at all.

PR #75, opened 2020-01-07
Provide flutter/flutter LUCI builder with fuchsia_ctl version

iskakaushik:


config/main.star
```diff
@@ -285,6 +285,7 @@ def linux_builder(properties = {}, caches=None, cores='8', category='Linux', os=
     properties = properties,
     caches = linux_caches,
     category = category,
+    fuchsia_ctl_version = 'version:0.0.10',
```
godofredoc: is it a property on linux_prod_builder(name='Linux|frwk' lines 336 and 339 a better place for this config? it would also make it consistent with the engine builders where it is passed as a property

iskakaushik: This should be equivalent to that. This would still pass it as properties on both the builders. And given that both the try-job and the scheduled-job would want to have the fuchsia tests.

Ideally I would move this to a place where both the engine and framework tryjobs will inherit it, but for now i think this should be ok.

dnfield:
When I run `main.star` with lucicfg 1.14, this fails.  Am I doing something wrong?

dnfield:
Hm.  I suspect you might have manually updated the .cfg file?

PR #82, opened 2020-01-24
builders for hotfixes

dnfield:
This should automatically build new commits to hotfix branches and fix https://github.com/flutter/flutter/issues/49381


config/main.star
```diff
@@ -138,7 +154,9 @@ def console_view(name, repo, refs = ['refs/heads/master'], exclude_ref = None):
 
 console_view('cocoon', COCOON_GIT)
 console_view('framework', FLUTTER_GIT)
+console_view('hotfix-framework', FLUTTER_GIT, [HOTFIX_REFS])
 console_view('engine', ENGINE_GIT)
```
godofredoc: Wouldn't this also include the hotfix_refs should we explicitly filter it?

config/main.star
```diff
@@ -99,12 +100,27 @@ luci.gitiles_poller(
...
+  name = 'hotfix-gitiles-trigger-framework',
+  bucket = 'prod',
+  repo = FLUTTER_GIT,
+  refs = [HOTFIX_REFS],
+)
+
 luci.gitiles_poller(
   name = 'master-gitiles-trigger-engine',
   bucket = 'prod',
   repo = ENGINE_GIT,
```
godofredoc: should this also have refs to filter out hotfixes-ref and do not trigger tasks twice?

config/main.star
dnfield: I don't think it does today but I don't mind being more explicit.

dnfield: Actually - it defaults to master: https://chromium.googlesource.com/infra/luci/luci-go/+/refs/heads/master/lucicfg/doc/README.md#luci.gitiles_poller

config/main.star
dnfield: Ditto - not specifying it makes it default ot master.

PR #86, opened 2020-01-29
Run yapf to format main.star file.

godofredoc:
https://github.com/flutter/flutter/issues/46782

yapf gives the file a nice standard format easy to follow along.


godofredoc:
yapf -i <filename> I download it from cipd, we can integrate it later when we move the recipes to its own repo.

PR #89, opened 2020-02-05
Update hotfix builders to use hotfix recipes

christopherfujino:
This is to support a hotfix to flutter stable v1.12.13. Currently changes to the engine recipe since v1.12.13 have broken re-building stable. This change will pin the hotfix builders to use the versions of the framework & engine recipes at the time v1.12.13 was released.

This is not a permanent fix; this change will be reverted after the successful hotfix release. The long-term fix will involve the Flutter team owning the LUCI build recipes.

### DIFF
--- a/config/main.star
+++ b/config/main.star
@@ -150,7 +150,9 @@ def recipe(name):
 
 recipe('flutter/cocoon')
 recipe('flutter/flutter')
+recipe('flutter/flutter_v1_12_13')
 recipe('flutter/engine')
+recipe('flutter/engine_v1_12_13')
 recipe('flutter/engine_builder')
 recipe('flutter/ios-usb-dependencies')
 recipe('flutter/web_engine')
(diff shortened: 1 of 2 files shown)


config/main.star
```diff
@@ -390,6 +392,7 @@ COMMON_HOTFIX_FRAMEWORK_BUILDER_ARGS = merge_dicts(
     COMMON_FRAMEWORK_BUILDER_ARGS, {
         'console_view_name':
         'hotfix-framework',
+        'recipe': 'flutter/flutter_v1_12_13', # TODO(fujino) revert back to 'flutter/flutter' after hotfix
```
dnfield: We can just specify this on the hotfix builders directly.

dnfield: e.g. somewhere around line 430, we can specify `recipe=` and similar for the other hotfix builders.

christopherfujino: I'm a python n00b:

```
linux_prod_builder(name='Linux hotfix|frwk',
                   properties={'shard': 'framework_tests'},
                   **COMMON_HOTFIX_FRAMEWORK_BUILDER_ARGS,
                   recipe='flutter/flutter_v1_12_13'),
```
will the recipe i specify override the one in `**COMMON_HOTFIX_FRAMEWORK_BUILDER_ARGS`?

christopherfujino: When I try this in a repl, I'm getting a typeError from python about multiple values for the same named argument. I think to override the original recipe, it has be done in a `merge_dict`.

dnfield: This isn't python, it's starlark :D  I'll come over.

dnfield: Ok.  I misread this.  This is fine.  But we don't need the todo - we won't want to bump this until we cut a new hotfix version.

christopherfujino: done.

PR #92, opened 2020-03-06
Add schedulers for the new release branches

digiter:
As release branches will be named `flutter-[x].[y]-candidate.[n]`, we need a new gitiles_poller that watch these release branches.

Bug: https://github.com/flutter/flutter/issues/52058


config/luci-scheduler.cfg
```diff
@@ -684,6 +684,14 @@ trigger: <
     refs: "regexp:refs/heads/master"
   >
 >
+trigger: <
+  id: "release-gitiles-trigger-framework"
+  acl_sets: "prod"
+  gitiles: <
+    repo: "https://chromium.googlesource.com/external/github.com/flutter/flutter"
+    refs: "regexp:refs/tags/flutter-\\d+\\.\\d+-candidate\\.\\d+"
+  >
```
godofredoc: The format of the branch we created for testing is vx.y.z we can change it to flutter-...candidate.. once we finished with the testing and we get confirmation from @tvolkert and @sosinski the final format.

tvolkert: Yep - as of now, the format will be `flutter-[x].[y]-candidate.[n]`

config/main.star
```diff
@@ -117,6 +117,14 @@ luci.gitiles_poller(
     refs=[HOTFIX_REFS],
 )
 
+
+luci.gitiles_poller(
+    name='release-gitiles-trigger-framework',
+    bucket='prod',
+    repo=FLUTTER_GIT,
+    refs=["refs/tags/flutter-\d+\.\d+-candidate\.\d+"],
```
godofredoc: Can we add this on the top as a constant, that will help with updating it when necessary?

config/luci-scheduler.cfg
godofredoc: Thanks, keeping this format here will ask @keyonghan to update the regex in the backend and @kf6gpe to create a new branch with this naming convention.

config/main.star

PR #93, opened 2020-03-13
Update Milo for release branches

digiter:
This will add a list_view to Milo, similar to hotfix-framework:
Zp97pjjE5vP

I think this is good enough for our use case, as test results will be added to release branches via GitHub webhook. Having a list_view per release branch would take more than enough screen space on the UI.

Bug: https://github.com/flutter/flutter/issues/52058


config/main.star
```diff
@@ -179,6 +179,7 @@ def console_view(name, repo, refs=['refs/heads/master'], exclude_ref=None):
 
 console_view('framework', FLUTTER_GIT)
 console_view('hotfix-framework', FLUTTER_GIT, [HOTFIX_REFS])
+console_view('release-framework', FLUTTER_GIT, [RELEASE_REFS])
```
godofredoc: release may be confusing as we have a release branch. Let's call it release-branches-framework.

config/main.star
```diff
@@ -408,6 +409,15 @@ COMMON_HOTFIX_FRAMEWORK_BUILDER_ARGS = merge_dicts(
...
     })
 
+COMMON_RELEASE_FRAMEWORK_BUILDER_ARGS = merge_dicts(
+    COMMON_FRAMEWORK_BUILDER_ARGS, {
+        'console_view_name': 'release-framework',
+        'triggered_by': ['release-gitiles-trigger-framework'],
+        'triggering_policy':
+        scheduler.greedy_batching(max_batch_size=1,
+                                  max_concurrent_invocations=3),
+    })
```
godofredoc: please run yapf formatting on this file.

config/main.star
```diff
@@ -408,6 +409,15 @@ COMMON_HOTFIX_FRAMEWORK_BUILDER_ARGS = merge_dicts(
                                   max_concurrent_invocations=3),
     })
 
+COMMON_RELEASE_FRAMEWORK_BUILDER_ARGS = merge_dicts(
+    COMMON_FRAMEWORK_BUILDER_ARGS, {
+        'console_view_name': 'release-framework',
```
godofredoc: Do we need to add an additional grouping level for branches?

config/main.star
digiter: How about "candidate-framework" as it's shorter and the branches are release candidates. /cc @tvolkert

digiter: @godofredoc This is the open question.

digiter: @pcsosinski

godofredoc: We definitely need the "framework" piece of the name as later on we'll need to add the same configs for the engine repo.

pcsosinski: candidate-framework lgtm

digiter: Changed to "candidate-framework".

config/main.star

config/main.star
digiter: I'm not sure how additional grouping works, from the discussion on luci users, there seems no good way to view lots of branches.

godofredoc: Let's add it in a follow up PR

PR #98, opened 2020-04-20
Revert "Update to use fuchsia_ctl 0.0.21."

godofredoc:
Reverts flutter/infra#97 we started getting ssh key problems.


godofredoc:
TBR

PR #102, opened 2020-04-23
Add builder configuration for flutter recipe bundler.

godofredoc:
Bug:


config/main.star
```diff
@@ -199,6 +206,14 @@ def recipe(name):
         cipd_version='refs/heads/master',
     )
 
+# Flutter recipe-bundler definitio
```
christopherfujino, line 209: spelling: `definitio -> definition`

christopherfujino:
Could you also clarify if there is a strategy for tracking recipe bundle versions for beta or stable cherry pick builds.

godofredoc:
@christopherfujino @digiter 

Added all the information to the bug:

https://github.com/flutter/flutter/issues/45630#issuecomment-618601154

Please let me know if we need more info.

config/main.star
godofredoc: Fixed!

config/main.star
```diff
@@ -457,6 +473,15 @@ COMMON_LINUX_COCOON_BUILDER_ARGS = {
...
 }
 
+COMMON_LINUX_RECIPES_BUILDER_ARGS = {
+    'recipe': 'recipe_bundler',
+    'console_view_name': 'recipes',
+    'triggered_by': ['master-gitiles-trigger-recipes'],
+    'triggering_policy':
+        scheduler.greedy_batching(
+            max_batch_size=1, max_concurrent_invocations=3),
+}
```
digiter, line 483: What is the location of the generated recipe bundle?

godofredoc: Good point, I didn't add the properties. Updating ...

godofredoc: Updated!

config/main.star
```diff
@@ -457,6 +473,23 @@ COMMON_LINUX_COCOON_BUILDER_ARGS = {
...
+    'recipe': 'recipe_bundler',
+    'console_view_name': 'recipes',
+    'triggered_by': ['master-gitiles-trigger-recipes'],
+    'triggering_policy':
+        scheduler.greedy_batching(
+            max_batch_size=1, max_concurrent_invocations=3),
+    'properties': {
+        'package_name_prefix': 'flutter/recipe_bundles',
+        'package_name_internal_prefix': 'flutter_internal/recipe_bundles',
+        'recipe_bundler_vers': 'git_revision:2ed88b2c854578b512e1c0486824175fe0d7aab6',
```
christopherfujino, line 486: So I'm guessing this revision is the reason we need an auto-roller that's referenced in the issue?

godofredoc: Not actually, we'll need to manually update that revision if the bundling process changes. The reason for the autoroller are the dependencies on other recipes. E.g. we will depend on depot_tools, recipe_engine, fuchsia_recipes and chromium recipes and ideally we should be using updated versions of those deps.

christopherfujino: Ahh, I see. Those will be rolled into the https://flutter.googlesource.com/recipes repo?

godofredoc: Yes, they will be using the versions from this file https://flutter.googlesource.com/recipes/+/refs/heads/master/infra/config/recipes.cfg to download the dependencies at bundling time.

PR #108, opened 2020-04-28
Configs to enable  unit test configurations.

godofredoc:
Bug:


config/lib/common.star
```diff
@@ -0,0 +1,201 @@
+# Copyright 2020 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+
+
+"""Code that's shared between turquoise and fuchsia."""
```
christopherfujino, line 6: Should redact sensitive comment

config/lib/common.star
```diff
@@ -0,0 +1,201 @@
...
+        https://foo.googlesource.com/bar/baz -> foo-bar-baz
+        https://foo.other-domain.com/bar/baz -> foo.other-domain.com-bar-baz
+    """
+    domain_and_path = repo_url.split("://")[1].split("/")
+    domain = domain_and_path[0]
+    path = domain_and_path[1:]
+    prefix = ""
+    if domain.endswith(".googlesource.com"):
+        prefix = domain[:-len(".googlesource.com")] + "-"
+    return prefix + "-".join(path)
```
christopherfujino, line 25: if `domain` is `"foo.other-domain.com"` won't prefix be `""` here?

config/main.star
```diff
@@ -124,6 +133,31 @@ luci.builder.defaults.properties.set({
     '\n84831b9409646a918e30573bab4c9c91346d8abd',
 })
 
+
+####################### Flutter Builder Definitions ##################################
```
digiter, line 137: Nit: explain the purpose of this builder in comment, for example: "this builder is for CQ attempt of recipe changes."

config/main.star
```diff
@@ -124,6 +133,31 @@ luci.builder.defaults.properties.set({
...
+common.cq_group(repos.FLUTTER_RECIPES)
+common.builder(
+    name="recipes-unittest-only",
+    builder_group=builder_groups.recipes_try,
+    # This builder is very quick to run, so we run it on every CQ attempt to
+    # minimize the chances of expectation file conflicts between CLs that land
+    # around the same time.
+    cq_disable_reuse=True,
+    executable=recipes.recipe(name="recipes"),
```
digiter, line 146: Nit: this line contains 3 "recipe", it's a bit hard to understand what each refers to. Maybe rename the 2nd "recipe" to "define" or "get".

config/lib/builder_groups.star
```diff
@@ -0,0 +1,45 @@
...
+        views=views,
+    )
+
+
+builder_groups = struct(
+    recipes_try=_group(
+        account=accounts.FLUTTER_TRY,
+        bucket="try",
+        pool="luci.flutter.try",
+	triggering_repos=(
```
mohrr, line 39: nit: indentation?

config/lib/common.star
mohrr: It looks like this was an issue where it was copied from too. I'm fixing it there.

config/main.star
godofredoc: > Please format with yapf.

Done

godofredoc: > Nit: this line contains 3 "recipe", it's a bit hard to understand what each refers to. Maybe rename the 2nd "recipe" to "define" or "get".
The naming is inline with the luci APIs and modules. In this case recipes is the module that will contain wrapper functions to create recipe configs. The second part is actually a wrapper function to create a recipe(luci.recipe) and name is the recipe file name(the name of this recipe is unfortunate but it will appear just twice in all the flutter configs).

config/main.star
godofredoc: Done

config/lib/common.star
godofredoc: fixed

config/lib/builder_groups.star
godofredoc: Done! fixed running yapf in all the files.

PR #109, opened 2020-04-29
Update autoroller builder to pass a flutter owned gcs bucket.

godofredoc:
The builder has been failing because it tries to upload metadata to a
chromium owned bucket.

Bug:


config/main.star
```diff
@@ -149,6 +149,27 @@ common.builder(
     },
     service_account=accounts.FLUTTER_TRY,
 )
+
+# Builder for recipes autoroller.
```
digiter, line 153: Nit: Builder as recipes autoroller. Since this builder is the autoroller.

godofredoc: Done!

godofredoc:
> LGTM, but I suggest putting the starlark and generated files in separate directories so it's easier to review. Mostly I review starlark closely and skim generated output but they're mixed together here.

Moved config files to generated/flutter

PR #111, opened 2020-04-30
Adjust permissions to allow manual triggering of tasks.

godofredoc:
This is required to enable retry functionality on the flutter dashboard.


config/main.star
```diff
@@ -64,18 +64,26 @@ luci.project(
             ],
             groups='all',
         ),
-        acl.entry(acl.BUILDBUCKET_TRIGGERER,
-                  groups='project-flutter-prod-schedulers'),
-        acl.entry(acl.BUILDBUCKET_TRIGGERER,
-                  users='luci-scheduler@appspot.gserviceaccount.com'),
```
mohrr, line 70: Did you intend to remove luci-scheduler's permission to trigger builds?

godofredoc: I added luci-scheduler@appspot.gserviceaccount.com to project-flutter-prod-schedulers. 

The plan is to keep only project-flutter-prod-schedulers with acl.BUILDBUCKET_TRIGGERER permissions. As it was  only admins and luci-scheduler@appspot.gserviceaccount.com were able to trigger tasks.

digiter: I'd like to know where is project-flutter-prod-schedulers configured?

PR #117, opened 2020-05-06
Add web engine prod builders.

godofredoc:
It also:

  * Moves repo constants to lib.
  * Moves common funtions to lib.

Bugs:
  https://github.com/flutter/flutter/issues/55973


config/lib/repos.star
```diff
@@ -3,4 +3,12 @@
 # found in the LICENSE file.
 
 # Frequently used git repository URLs.
-repos = struct(FLUTTER_RECIPES="https://flutter.googlesource.com/recipes", )
+repos = struct(
```
christopherfujino, line 6: Just a question, is `struct` a built-in to python, starlark, or is it a lucicfg function?

godofredoc: It is a function of starlark language: https://docs.bazel.build/versions/3.1.0/skylark/lib/struct.html

PR #121, opened 2020-05-07
update builders for 1.17.0 stable release

christopherfujino:
Update refs regex and recipe to use for stable builders and delete the now deprecated hotfix builders.


christopherfujino:
Hmm, `validate_config` says that:

```
Error: the following files need to be regenerated: luci/commit-queue.cfg.
  Run `lucicfg generate "./config/main.star"` to update them.
```

when running that I get no diff on `commit-queue.cfg` though, and I don't see why there should be...

christopherfujino:
> This PR has no diff on that file and when running locally I still get no diff on `commit-queue.cfg` though. It also seem to me that given my change, there shouldn't be...

Okay, this is a problem in master, will fix separately...

PR #127, opened 2020-05-15
[fuchsia_ctl] Add a try builder

digiter:
Bug: https://github.com/flutter/flutter/issues/56138


config/main.star
```diff
@@ -396,6 +397,10 @@ luci.list_view(
     name='engine-try',
     title='Engine try builders',
 )
+luci.list_view(
```
godofredoc, line 400: This will be only for try jobs. Let's not create a view for it.

config/main.star
```diff
@@ -975,3 +980,8 @@ ios_tools_builder(name='usbmuxd|usbmd')
 ios_tools_builder(name='openssl|ssl')
 ios_tools_builder(name='ios-deploy|deploy')
 ios_tools_builder(name='libzip|zip')
+
+common.linux_try_builder(name='fuchsia_ctl|fctl',
+                         recipe='fuchsia_ctl',
+                         list_view_name='fuchsia_ctl-try',
+                         properties={'fuchsia_ctl_version': None})
```
godofredoc, line 987: Any reason we need to pass this property as none?

config/main.star
digiter: Removing list_view will cause a bigger change in `common._try_builder`, I will do that in follow up PRs.

config/main.star
digiter: Having fuchsia_ctl_version is confusing because the recipe ignores this property. And we don't want to specify the fuchsia_ctl_version because we'd like to test the tip of master.

godofredoc: ok, not a blocker but something we may want not want to force in the future.

PR #128, opened 2020-05-16
Refactor recipes and framework builders

digiter:
To have better readability, this no-op change refactors `main.star`.
1. Creates a global BRANCHES dict to encapsulate all branches metadata.
1. Creates `recipes_row` and `framework_rows` functions to encapsulate the builders, console views, list views, triggers, recipes etc. So that relevant information are gathered in a small scope rather than the whole file.
1. Replaced global constants like COMMON_STABLE_FRAMEWORK_BUILDER_ARGS with local arguments.
1. Generates builders with for loops over branches.
1. A few drive-by format changes.

If this looks good, we could continue refactoring engine builders with this pattern.


config/main.star
```diff
@@ -124,50 +151,229 @@ luci.builder.defaults.properties.set({
     '\n84831b9409646a918e30573bab4c9c91346d8abd',
 })
 
+############################ End Global Defaults ############################
+
+
+def recipes_row():
```
godofredoc, line 157: Here and below can we raname recipes_row to something like recipes_configs()?

config/main.star
```diff
@@ -124,50 +151,229 @@ luci.builder.defaults.properties.set({
...
+            'repo_specs': [
+                'flutter.googlesource.com/recipes=FETCH_HEAD,refs/heads/master',
+            ],
+        }.items(),
+        console_category=console_name,
+        console_short_name='bdlr',
+    )
+
+
+recipes_row()
```
godofredoc, line 232: Now that we have functions we can actually move all of the above functions to lib/recipe_configs.star import it in the top and just call recipes_config() here.

config/main.star
```diff
@@ -124,50 +151,229 @@ luci.builder.defaults.properties.set({
...
+    )
+    common.windows_try_builder(
+        name='Windows|frwk',
+        recipe=branch_args['master']['recipe'],
+        list_view_name=list_view_name,
+        **platfrom_args['windows'],
+    )
+
+
+framework_rows()
```
godofredoc, line 354: Same comment as above, let's create a configs library per repo.

config/main.star

config/main.star

config/main.star

PR #129, opened 2020-05-27
temporarily disable "Mac Web Engine" until flake is fixed

christopherfujino:
The flake is being tracked in https://github.com/flutter/flutter/issues/58012.


christopherfujino:
Closing this as this task should be skipped from cocoon, rather than the builder config.

digiter:
I'm not sure about this, how about letting cocoon to stop sending the build result of "Mac Web Engine" to github?

christopherfujino:
> I'm not sure about this, how about letting cocoon to stop sending the build result of "Mac Web Engine" to github?

Won't we still get 500s if we do that?

digiter:
We can refactor `luci_status.dart` to avoid 500s by making the file understand how to ignore a builder result.

christopherfujino:
> We can refactor `luci_status.dart` to avoid 500s by making the file understand how to ignore a builder result.

This sounds like the correct long-term fix. I don't have the time to figure out how to do implement that, however. I opened this PR as a quick band-aid to re-open the engine tree. Does someone have the capacity to make this update? @digiter

christopherfujino:
@digiter should your suggestions be part of https://github.com/flutter/flutter/issues/57155, or tracked separately?

digiter:
I think it can be an additional item to flutter/flutter#57155.

christopherfujino:
> I think it can be an additional item to [flutter/flutter#57155](https://github.com/flutter/flutter/issues/57155).

SGTM, i'll update it.

PR #130, opened 2020-05-29
re-enable mac web engine

christopherfujino:
Blocked by pending recipe update. @nturgut is trying to debug https://github.com/flutter/flutter/issues/58012, but is not able to run the recipe with LED because the builder is gone. After the recipe is updated to skip the flaky step, this PR will be landed to re-enable the builder, and Nurhan will be able to debug the flake with LED.

This PR reverses https://github.com/flutter/infra/pull/129


nturgut:
I merged the recipe change.

nturgut:
Thanks for merging the PR.

This is an example run of how Mac Wen Engine will look like: https://chromium-swarm.appspot.com/task?id=4c7ba49f6bd25110

I think it is still better than nothing :) Even with no unit tests, we can keep validating the build. I will update the recipe, when I resolved the flakes.

PR #131, opened 2020-06-02
Do not use osx from caches.

godofredoc:
This is generating a lot of flakiness on mac builders. Removing it from
the caches won't have a huge impact because it is still cached at the
cipd level.

Bug:
  https://github.com/flutter/flutter/issues/58516


godofredoc:
> RSLGTM. just wondering, is there another underlying problem here why the caches are so flaky? if the cache wasn't buying us much, maybe it's not worth fixing?

Probably there are more problems. The reason I believe it is more flaky on mac is because we have less machines and the caches are hit more frequently. I'll compare the before and after to see the actual implication of removing the cache but probably is not going to be a major problem as CIPD packages are already cached.

PR #136, opened 2020-06-16
Move engine configurations to their own file.

godofredoc:
Flutter configurations became very complex with a lot of duplication.
This pr refactors the engine builder configs as functions to avoid
duplication and improve readability.

Bug:


config/lib/repos.star
```diff
@@ -6,9 +6,11 @@
 repos = struct(
     FLUTTER_RECIPES="https://flutter.googlesource.com/recipes",
     COCOON=
-    'https://chromium.googlesource.com/external/github.com/flutter/cocoon',
+    'https://flutter.googlesource.com/flutter/cocoon',
```
digiter, line 9: This url doesn't exist.

config/engine_config.star
```diff
@@ -0,0 +1,471 @@
...
+to take advantage of led recipe builder tests.
+"""
+
+load("//lib/common.star", "common")
+load("//lib/repos.star", "repos")
+
+FUCHSIA_CTL_VERSION = 'version:0.0.23'
+
+
+def _setup(BRANCHES):
```
digiter, line 22: BRANCHES should be lower cased since it's not a global variable. The one in framework_config should also be fixed.

config/engine_config.star
```diff
@@ -0,0 +1,471 @@
...
+
+    Returns:
+      A string with the recipe's full name.
+    """
+    if recipe_name == 'engine':
+        return 'engine_' + version if version else 'engine'
+    elif recipe_name == 'web_engine':
+        return 'web_engine_' + version if version else 'web_engine'
+    elif recipe_name == 'engine_builder':
+        return 'engine_builder_' + version if version else 'engine_builder'
```
digiter, line 60: Nit: so it seems this function could be simplified to `return recipe_name + ('_' + version if version else '')`

config/engine_config.star
```diff
@@ -0,0 +1,471 @@
...
+    luci.recipe(
+        name=full_recipe_name('web_engine', version),
+        cipd_package='flutter/recipe_bundles/flutter.googlesource.com/recipes',
+        cipd_version='refs/heads/master',
+    )
+    luci.recipe(
+        name=full_recipe_name('engine_builder', version),
+        cipd_package='flutter/recipe_bundles/flutter.googlesource.com/recipes',
+        cipd_version='refs/heads/master',
+    )
```
digiter, line 79: Nit: consider a loop for these three.

config/lib/repos.star
christopherfujino: Is there a tracking issue or doc about migrating these repos to flutter.googlesource.com?

config/engine_config.star
```diff
@@ -0,0 +1,471 @@
+#!/usr/bin/env lucicfg
+# Copyright 2020 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+"""
+Configurations for the flutter engine repository.
+
+The schedulers pull commits indirectly from GoB repo
+(https://chromium.googlesource.com/external/github.com/flutter/engine)
```
christopherfujino, line 9: should this be updated to flutter.googlesource.com/flutter/engine?

config/engine_config.star
```diff
@@ -0,0 +1,471 @@
...
+which is mirrored from https://github.com/flutter/engine.
+
+Try jobs use github directly but are also registered with luci-cq
+to take advantage of led recipe builder tests.
```
christopherfujino, line 13: what are led recipe builder tests?

config/engine_config.star
```diff
@@ -0,0 +1,471 @@
...
+        'linux': {},
+        'mac': {
+            'caches':
+            [swarming.cache(name='flutter_cocoapods', path='cocoapods')],
+        },
+        'windows': {},
+    }
+
+    for branch in BRANCHES:
+        engine_recipes(BRANCHES[branch]['version'], )
```
christopherfujino, line 34: nit: no comma

config/engine_config.star
```diff
@@ -0,0 +1,471 @@
...
+    if recipe_name == 'engine':
+        return 'engine_' + version if version else 'engine'
+    elif recipe_name == 'web_engine':
+        return 'web_engine_' + version if version else 'web_engine'
+    elif recipe_name == 'engine_builder':
+        return 'engine_builder_' + version if version else 'engine_builder'
+
+
+def engine_recipes(version):
+    """Creates a luci recipes for a given code version."""
```
christopherfujino, line 64: nit: sp `recipes` -> `recipe`

christopherfujino, line 64:
```suggestion
    """Creates a luci recipe for a given code version."""
```

config/engine_config.star
christopherfujino, line 34:
```suggestion
        engine_recipes(BRANCHES[branch]['version'])
```

config/engine_config.star
```diff
@@ -0,0 +1,471 @@
...
+        'build_android_vulkan': build_android_vulkan,
+        'build_ios': build_ios,
+        'build_android_jit_release': build_android_jit_release,
+    }
+    if (build_ios):
+        properties['ios_debug'] = ios_debug
+        properties['ios_profile'] = ios_profile
+        properties['ios_release'] = ios_release
+        properties['no_bitcode'] = no_bitcode
+    if (build_ios):
```
christopherfujino, line 109: is this duplicated from the above 5 lines?

config/lib/repos.star
godofredoc: This is expected, they are used for fake configs that provide the list of CQ builders even though we are not using LUCI CQ and they don't like github configs.

config/engine_config.star
godofredoc: Done

config/engine_config.star
godofredoc: Done

config/engine_config.star
godofredoc: Done

config/engine_config.star
godofredoc: Actually chromium.googlesource is correct. flutter.googlesource.com is used only because we we need fake cq configs to allow the presubmit tests to detect configured builders.

config/engine_config.star
godofredoc: Those are tests that run the real try builders affected by a change to the recipes. E.g. if you change something affecting engine.py then it will run all win, mac, and linux builders that use the recipe as presubmit.

config/engine_config.star
godofredoc: Done.

config/engine_config.star
godofredoc: yes, nice catch I just copy pasted everything from main.star. Removed the duplicated lines.

config/engine_config.star
godofredoc: Done!

config/lib/repos.star
christopherfujino: Could you add a comment?

godofredoc:
> LGTM with a nit to add a comment to `config/lib/repos.star` about the fake repo stubs

Actually I got the repo urls back to the chromium external because it would've affected the prod pollers and modified the cq_groups to replace "." with "_" in the name generated from the url.

PR #137, opened 2020-06-17
Update recipe-deps-roller CIPD package

orn688:
This version includes http://crrev.com/c/2247497, which causes the
recipe roller to respect the new `no_cc_authors` recipes.cfg field so it
won't spammily CC authors of rolled recipe changes.


orn688:
Hi @godofredoc - is this the correct place to submit changes like this, or do you use Gerrit?

godofredoc:
> Hi @godofredoc - is this the correct place to submit changes like this, or do you use Gerrit?

This is the correct place for flutter recipe config changes. Let me check why the check is failing.

godofredoc:
The check was failing because you don't have write access to the repo and the secrets were not decoded. Re-running the checks to ensure it is safe to land.

PR #139, opened 2020-06-22
Update recipe-deps-roller CIPD package

orn688:
This version includes http://crrev.com/c/2256318, which is a reland of a
reverted CL that makes it safe to call `recipes.py dump_specs` in repos
where the command isn't supported yet.

Bug: fxbug.dev/54380


orn688:
@godofredoc mind taking a look at this again?

PR #143, opened 2020-06-29
Complete refactoring of recipe configurations.

godofredoc:
Bug:


config/iostools_config.star
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env lucicfg
+# Copyright 2020 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+"""
+Configurations for ios_tool builders.
+
+The schedulers pull commits indirectly from GoB repo (https://chromium.googlesource.com/external/github.com/flutter/flutter)
```
christopherfujino, line 8: commits actually come from https://flutter-mirrors.googlesource.com/

config/cocoon_config.star
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env lucicfg
+# Copyright 2019 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+"""
+Configurations for the cocoon repository.
+
+The schedulers pull commits indirectly from GoB repo (https://chromium.googlesource.com/external/github.com/flutter/flutter)
```
christopherfujino, line 8: pulls from https://chromium.googlesource.com/external/github.com/flutter/cocoon

config/packaging_config.star
```diff
@@ -0,0 +1,114 @@
...
+        cipd_package='flutter/recipe_bundles/flutter.googlesource.com/recipes',
+        cipd_version='refs/heads/master',
+    )
+
+
+def packaging_prod_config(platform_args, branch, version, ref):
+
+    # Defines console views for prod builders
+    console_view_name = ('packaging'
+                         if branch == 'master' else '%s_packaging' % branch)
```
christopherfujino, line 63: this function should never get called with `branch == 'master'`, right?

christopherfujino: Should we throw here (or somehow escalate a warning in a way we'll see) if this function ever gets invoked on the master branch? I'm worried a future refactor will lead to us building packages for every single master commit. At the minimum, we should exit early.

godofredoc: Added an explanation and added functionality to fail if we ever try to build from master.

config/cocoon_config.star
godofredoc: Updated

config/iostools_config.star
godofredoc: Updated.

godofredoc:
> 2 nits about comments and then a question about ensuring we don't regress in the future and build packages on master.

Updated the PR addressing the comments.

PR #144, opened 2020-06-30
Optimize execution of led recipe tests.

godofredoc:
Recipe's auto-roller tries to roll dependencies from chromium/build,
depot_tools, recipe_engine and fuchsia every hour.

This PR is adding a new parameter to try_builders to decide if they
will be added to cq configuration file and which builders will be
considered by the recipe's presubmit tests.

It also keeps a single mac builder per repository and changes the
scheduler to run twice per day at 6AM PST and 7PM PST in preparation to
remove the experimental bit from led recipe tests.

Bug:


config/generated/flutter/luci/commit-queue.cfg
```diff
@@ -53,9 +53,6 @@ config_groups {
       builders {
         name: "flutter/try/Linux Android Debug Engine"
       }
-      builders {
```
christopherfujino, line 56: Are all of these builders that were removed supposed to be removed? For example, "Linux Engine Drone" was removed, but not "Windows Engine Drone"?

godofredoc: Yes, drone builders should not run by themselves and tried to keep only one mac builder for engine led tests.

PR #146, opened 2020-07-01
Enable formatter and linter for recipe configs.

godofredoc:
Bug:


godofredoc:
> The change LGTM. How does the formatter work? will lucicfg auto-format from now on, or do we need to run a special command?

I'll create a new PR documenting the new structure and providing guidelines for new changes. The current change process needs to include the following commands:

```
cd //config/
lucicfg fmt .
lucicfg lint .
git add ...
```

The current presubmit check will fail if there are any formatting or lint problems.

christopherfujino:
Ahh, ok, if presubmit validates it then I'll figure it out :)

PR #147, opened 2020-07-01
fix gitiles poller trigger for packaging builders

christopherfujino:
1. Modified the `BRANCHES` dict to have two refs, `testing-ref` and `release-ref`.
2. Modify the packaging pollers to use `release-ref`, the rest to use `testing-ref`
3. Rename the packaging pollers to have `packaging` in the name, since I was running into lucicfg errors that look like `Error: luci.gitiles_poller("prod/stable-gitiles-trigger-framework") is redeclared, previous declaration:` and I didn't want to refactor further to share the pollers. This could be refactored later.

### DIFF
--- a/config/engine_config.star
+++ b/config/engine_config.star
@@ -34,7 +34,7 @@ def _setup(branches):
             platform_args,
             branch,
             branches[branch]["version"],
-            branches[branch]["ref"],
+            branches[branch]["testing-ref"],
         )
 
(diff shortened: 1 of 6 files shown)


christopherfujino:
wait, I need to fix a bug

christopherfujino:
Actually my fix didn't change the generated configs, so I guess it was ok.

PR #149, opened 2020-07-08
remove explicit max_batch_size, should default to 1000

christopherfujino:
The [default for this value is 1000](https://chromium.googlesource.com/infra/luci/luci-go/+/master/scheduler/appengine/messages/config.proto#130). I was overly conservative when I set this, and as a result a lot of unnecessary commits get built when I update one of these dependencies.


christopherfujino:
> I like that changes are now very encapsulated.

Yeah, it's great that what feels like a 1-liner change...is :)

PR #154, opened 2020-07-14
make beta regex more general for experiment

christopherfujino:
@pcsosinski will be conducting an experiment, so we need to make the branch regex more general so that it will pick up a different 1.20.x RC branch.


pcsosinski:
do we need the framework builders or just the engine?

christopherfujino:
> do we need the framework builders or just the engine?

We don't need the framework builders, but this regex is shared between both, and it we can safely ignore the builds.

PR #156, opened 2020-07-14
Use a subset of machines for ios tests.

godofredoc:
There is a subset of machines that are failing ios tasks less frequently
than all the others. We are updating configurations to run iOS tasks
only on those machines.

Bug:
  https://github.com/flutter/flutter/issues/61267


config/lib/common.star
```diff
@@ -235,12 +235,20 @@ def _flutter_builder(
...
         cores = None,
+        dimensions = None,
         **kwargs):
-    dimensions = {
-        "pool": pool,
-        "cpu": "x64",
-        "os": os,
-    }
+    if dimensions:
+        dimensions = helpers.merge_dicts(dimensions, {
```
christopherfujino, line 241: shouldn't the order of these args be switched?

godofredoc: Done

godofredoc:
> I don't really know what's up with the older mac minis and how it could be causing app crashes on the simulator... I don't think we really have a good handle on the underlying cause of the test failures.
> However, whatever the infra team wants to do here is fine with me.
> LGTM once @christopherfujino is happy with the code.

If this change reduces the number of flakes then we'll know for sure is something related to the machines in 1808 and we can focus on finding which provisioning difference is affecting the ios tasks.

PR #159, opened 2020-07-17
Move led recipe tests out of experimental.

godofredoc:
This tests are intended to validate all the applicable builders against
recipe changes.

Bug:
  https://github.com/flutter/flutter/issues/61680


christopherfujino:
So we're removing these because they've been failing because of the bbagent issue?

godofredoc:
> So we're removing these temporarily because they've been failing because of the bbagent issue?

Yes that is correct, chromium team has already a fix for piping logs to power shell. Once that gets fixed I'll move all the builders to use bb_agent and reenable the windows builder for recipe's led tests.

christopherfujino:
> Yes that is correct, chromium team has already a fix for piping logs to power shell. Once that gets fixed I'll move all the builders to use bb_agent and reenable the windows builder for recipe's led tests.

that's great about the powershell fix, let's ship it.

PR #161, opened 2020-07-24
windows bots are failing for goldens

nturgut:
windows bots are failing for goldens


nturgut:
Thanks for the quick review!

PR #167, opened 2020-08-05
Add flutter sdk drone configurations.

godofredoc:
This will enable the drone builder in buildbucket to be able to test the
final recipe.

Bug:
  https://github.com/flutter/flutter/issues/62811


christopherfujino:
@godofredoc validation errors

christopherfujino:
So the idea is the `flutter/flutter` recipe will spawn builds of `flutter/flutter_drone`?

godofredoc:
> @godofredoc validation errors

fixed

godofredoc:
> So the idea is a single `flutter/flutter` build will spawn (and collect) multiple builds of `flutter/flutter_drone`?

Yes, from the dashboard point of view we'll have a single framework_test but the build will spawn 4 drones to run the tests. This way on high availability of bots the test will complete in a fraction of time.

PR #177, opened 2020-08-13
use latest recipes on beta

christopherfujino:


pcsosinski:
thanks.  a branch point for 1.21 is "imminent" as well.

PR #178, opened 2020-08-13
Skip LTO on try builders

dnfield:
Needs https://flutter-review.googlesource.com/c/recipes/+/5640

/cc @chinmaygarde @jason-simmons


dnfield:
Here's an example of a build where LTO added about 8 minutes: https://logs.chromium.org/logs/flutter/buildbucket/cr-buildbucket.appspot.com/8872059819503845344/+/steps/build_ios_debug/0/stdout

PR #181, opened 2020-08-14
Add a pre-submit builder for devicelab tests.

digiter:
We will add some devicelab tests running on LUCI in presubmit.

Bug: https://github.com/flutter/flutter/issues/53789


config/framework_config.star
```diff
@@ -419,6 +424,12 @@ def framework_try_config(platform_args):
             swarming.cache(name = "android_sdk", path = "android29"),
         ],
     )
+    common.linux_try_builder(
+        name = "testonly_devicelab_tests|tst_tests",
```
godofredoc, line 428: Do we need properties?

digiter: We don't need. But in order to remove them, we will have to perform a refactor to remove the defaults https://github.com/flutter/infra/blob/master/config/main.star#L158.

PR #183, opened 2020-08-15
Fix tests not showing up in milo.

godofredoc:
Bug:
  https://github.com/flutter/flutter/issues/63807


config/framework_config.star
```diff
@@ -229,8 +229,9 @@ def framework_prod_config(platform_args, branch, version, ref):
     common.linux_prod_builder(
         name = "Linux%s analyze|anlz" % ("" if branch == "master" else " " + branch),
         recipe = new_recipe_name,
-        console_view_name = None,
-        no_notify = True,
```
christopherfujino, line 233: I see you're removing `no_notify`, and it looks like it defaults to `False`. Is that ok?

godofredoc: Yes, this is because I copied the drone builder for all the new builders and drone is the only one with console_view as none and not_notify as true. Later when the builders are stable we'll need to configure the notifications to send emails when a builder fails.

PR #191, opened 2020-08-18
pin beta builder to recipe 1_21_0

christopherfujino:
Fixes https://github.com/flutter/flutter/issues/64078

Must land after https://flutter-review.googlesource.com/c/recipes/+/5842


christopherfujino:
> LGTM.
> 
> Btw, looks like there is a validation error.

thanks. forgot to generate...

PR #194, opened 2020-08-19
Make engine:master build with max_batch_size 1

dnfield:
Make sure we build every commit of the engine on master.

### DIFF
--- a/config/engine_config.star
+++ b/config/engine_config.star
@@ -158,7 +158,7 @@ def engine_prod_config(platform_args, branch, version, ref, fuchsia_ctl_version)
     if branch == "master":
         triggering_policy = scheduler.greedy_batching(
             max_batch_size = 1,
-            max_concurrent_invocations = 6,
+            max_concurrent_invocations = 3,
         )
     else:
         triggering_policy = scheduler.greedy_batching(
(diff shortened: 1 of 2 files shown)


config/engine_config.star
```diff
@@ -157,6 +157,7 @@ def engine_prod_config(platform_args, branch, version, ref, fuchsia_ctl_version)
     # Defines triggering policy
     if branch == "master":
         triggering_policy = scheduler.greedy_batching(
+            max_batch_size = 1,
             max_concurrent_invocations = 6,
```
godofredoc, line 161: Can we decrease the max_concurrent invocations to 3? this will use most of the current capacity if several commits land together. Actually if they land within 15mins as the replica will fetch them in batches

dnfield: Keeping engine:master up to date is probably the most critical thing we do with infra. The only place we shoul dbe constrained on this is on macOS, where we're about to get more machines.

If anything, I think this value should be higher, not lower.

godofredoc: If we submit as it is with no further planning on capacity we will be impacting negatively the SDK(flutter prod is shared with sdk and engine) and we can expect very long delays from ~1:00 to 5:00.

godofredoc: I'd suggest to adjust to 3, monitor, get capacity and then increase.

PR #197, opened 2020-09-02
Add LUCI config file for plugins

keyonghan:
This is part of https://github.com/flutter/flutter/issues/56143#issuecomment-685843581, adding configs for windows try builder.

### DIFF
--- a/config/plugins_config.star
+++ b/config/plugins_config.star
@@ -31,13 +31,13 @@ def plugins_try_config(platform_args):
     """Detailed builder configures for different platforms.
     
     Args:
-        platform_args (dict) - platform arguments.
-        For example: 
-        {
-            "windows": {
-                "caches": [swarming.cache(name = "pub_cache", path = ".pub-cache")],
+        platform_args: The platform arguments passed to luci builders.
+            For example: 
+            {
+                "windows": {
+                    "caches": [swarming.cache(name = "pub_cache", path = ".pub-cache")],
+                }
             }
-        }
     """
     # Defines a list view for try builders
     list_view_name = "plugins-try"
--- a/config/plugins_config.star
+++ b/config/plugins_config.star
@@ -29,16 +29,17 @@ def plugins_define_recipes():
 
 def plugins_try_config(platform_args):
     """Detailed builder configures for different platforms.
-    
+
     Args:
         platform_args (dict): The platform arguments passed to luci builders.
-            For example: 
+            For example:
             {
                 "windows": {
                     "caches": [swarming.cache(name = "pub_cache", path = ".pub-cache")],
                 }
             }
     """
+
     # Defines a list view for try builders
     list_view_name = "plugins-try"
     luci.list_view(
--- a/config/plugins_config.star
+++ b/config/plugins_config.star
@@ -30,7 +30,7 @@ def plugins_define_recipes():
 # Detailed builder configures for different platforms.
 #
 # [platform_args] has map structure, with platforms as keys.
-# Example:  
+# Example:
 # {
 #    "windows": {
 #        "caches": [swarming.cache(name = "pub_cache", path = ".pub-cache")],
@@ -50,7 +50,7 @@ def plugins_try_config(platform_args):
         recipe = "plugins/plugins",
         list_view_name = list_view_name,
         repo = repos.PLUGINS,
-        **platform_args["windows"],
+        **platform_args["windows"]
     )
 
 plugins_config = struct(setup = _setup)
--- a/config/plugins_config.star
+++ b/config/plugins_config.star
@@ -9,6 +9,7 @@ Configurations for the plugins repository.
 load("//lib/common.star", "common")
 load("//lib/repos.star", "repos")
 
+# Set default configurations for builders, and setup recipes.
 def _setup():
     platform_args = {
         "windows": {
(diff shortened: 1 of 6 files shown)


config/plugins_config.star
```diff
@@ -0,0 +1,49 @@
...
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
+"""
+Configurations for the plugins repository.
+"""
+
+load("//lib/common.star", "common")
+load("//lib/repos.star", "repos")
+
+def _setup():
```
godofredoc, line 12: Docs?

config/plugins_config.star
```diff
@@ -0,0 +1,49 @@
...
+    platform_args = {
+        "windows": {
+            "caches": [swarming.cache(name = "pub_cache", path = ".pub-cache")],
+        },
+    }
+    plugins_define_recipes()
+    plugins_try_config(platform_args)
+
+def plugins_define_recipes():
+    # Defines recipes
```
godofredoc, line 22: Add this line as doc and add more info?

config/plugins_config.star
```diff
@@ -0,0 +1,49 @@
...
+    luci.recipe(
+        name = "plugins/plugins",
+        cipd_package = "flutter/recipe_bundles/flutter.googlesource.com/recipes",
+        cipd_version = "refs/heads/master",
+    )
+
+def plugins_try_config(platform_args):
+    # Defines a list view for try builders
```
godofredoc, line 30: Same as above, please update as a doc and document the argument with an example.

config/plugins_config.star
```diff
@@ -0,0 +1,49 @@
...
+    # Defines plugins Windows platform try builders
+    common.windows_try_builder(
+        name = "Windows Plugins|windows",
+        recipe = "plugins/plugins",
+        list_view_name = list_view_name,
+        repo = repos.PLUGINS,
+        add_cq = True,
+        caches = [
+            swarming.cache(name = "pub_cache", path = ".pub_cache"),
+        ],
```
godofredoc, line 46: Use platform args here or remove the parameter and definition at the top?

config/plugins_config.star
```diff
@@ -0,0 +1,49 @@
...
+        title = "Plugins try builders",
+    )
+
+    # Defines plugins Windows platform try builders
+    common.windows_try_builder(
+        name = "Windows Plugins|windows",
+        recipe = "plugins/plugins",
+        list_view_name = list_view_name,
+        repo = repos.PLUGINS,
+        add_cq = True,
```
godofredoc, line 43: Do we want this builder to block submissions in the recipes repo?

keyonghan: What is the context about add_cq property, or any doc? I see builders from all other repos have this property set as `True`.

config/plugins_config.star
keyonghan: Changed to use args.

config/plugins_config.star
keyonghan: Added.

config/plugins_config.star
keyonghan: Done.

config/plugins_config.star
keyonghan: Removing it for now.

godofredoc: This is used to decide whether or not to add the builder to the commit_queue.cfg file. commit_queue.cfg is used to decide which builders will run on when a recipe is affected. E.g. if there is a change to the plugins/plugins recipe and add_cq=True then this builder will be run on that recipe change. Setting this value to True is the correct path to follow once we validate the builder runs correcty.

config/plugins_config.star
```diff
@@ -18,14 +19,23 @@ def _setup():
...
 
+# Detailed builder configures for different platforms.
+#
+# [platform_args] has map structure, with platforms as keys.
+# Example:  
+# {
+#    "windows": {
+#        "caches": [swarming.cache(name = "pub_cache", path = ".pub-cache")],
+#    }
+# }
```
godofredoc, line 38: rm if not used.

keyonghan: We are using `platform_args` now (`**platform_args["windows"]`).

godofredoc: I see, then if this is the doc for the function please move it inside the function.

config/plugins_config.star
keyonghan: Ahh, okay. Updated all docs to python format.

config/plugins_config.star
keyonghan: Done.

config/plugins_config.star
```diff
@@ -0,0 +1,58 @@
...
+        cipd_package = "flutter/recipe_bundles/flutter.googlesource.com/recipes",
+        cipd_version = "refs/heads/master",
+    )
+
+def plugins_try_config(platform_args):
+    """Detailed builder configures for different platforms.
+    
+    Args:
+        platform_args (dict) - platform arguments.
+        For example:
```
godofredoc, line 35: platform_args: The platform arguments passed to luci builder ...
   line continuation is indented to not confuse with other arguments.

keyonghan: Done. We don't need to specify type here?

keyonghan: I suppose the type as missed in the comment. Just added it back.

godofredoc (approved):

PR #198, opened 2020-09-02
fix typo in usbmuxd repo remote url

christopherfujino:
This URL is wrong, it should point to: https://flutter-mirrors.googlesource.com/libusbmuxd


jmagman (approved):
LGTM

PR #201, opened 2020-09-08
Update default os to Windows-Server for plugins windows tests

keyonghan:
Plugins repo is now supporting LUCI pre-submit windows test. But existing recipe is using default `Windows-10`, which fails the `flutter doctor` step: 

```
[X] Visual Studio - develop for Windows
    X Visual Studio not installed; this is necessary for Windows development.
```

This PR updates the os to `Windows-Server`, which works: 
https://logs.chromium.org/logs/flutter/led/stuartmorgan_google.com/8262c62e8def8247edb8e1e393a0f00077b60a9555dc8c7259f30762466019fb/+/steps/prepare_environment/0/steps/flutter_doctor/0/stdout

Related issue: 
https://github.com/flutter/flutter/issues/56143#issuecomment-687532941

### DIFF
--- a/config/generated/flutter/luci/cr-buildbucket.cfg
+++ b/config/generated/flutter/luci/cr-buildbucket.cfg
@@ -8489,7 +8489,7 @@ buckets {
       swarming_host: "chromium-swarm.appspot.com"
       dimensions: "cores:8"
       dimensions: "cpu:x64"
-      dimensions: "os:Windows-10"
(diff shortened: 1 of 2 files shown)


stuartmorgan (commented):
I guess I read the comments in the issue wrong; I thought the earlier runs were using -10 and it was switched to -Server later. LGTM if Windows-Server is what I was using in the run that worked :)

PR #202, opened 2020-09-08
Update cores to 32 for windows test in Plugins repo

keyonghan:
This is a follow-up of https://github.com/flutter/infra/pull/201.

The default cores for windows builders are 8, however the cores for os `Windows-Server` bot are 32. We need to update to 32 to allow LUCI find corresponding bots.


stuartmorgan (approved):
RSLGTM

PR #204, opened 2020-09-09
Fix tool tests mac builder.

godofredoc:
Even though the builder had all the configs correct it was using the
linux function instead of the mac one.

Bug:
  https://github.com/flutter/flutter/issues/65223


keyonghan (approved):
LGTM

PR #211, opened 2020-09-11
Add tricium to infra configs.

godofredoc:
This is required to enable tricium validations in the flutter recipes
repository.


keyonghan (approved):
LGTM

PR #212, opened 2020-09-11
Move tricium prod to the generated folder.

godofredoc:
Only the generated folder is being read by luci configurations.


digiter (approved):
LGTM

PR #214, opened 2020-09-12
Add cocoapod configurations.

godofredoc:
It also removes several properties not being used.

Bug:
  https://github.com/flutter/flutter/issues/65223


keyonghan (approved):
LGTM

christopherfujino (commented):
Should we use a gemfile? (this comment should have been on the recipe CL, but I just thought of it now). With rubygems, if you don't have a `gemfile.lock` file, transitive deps can change, and this broke us once with `fastlane`.

godofredoc:
> Should we use a gemfile? (this comment should have been on the recipe CL, but I just thought of it now). With rubygems, if you don't have a `gemfile.lock` file, transitive deps can change, and this broke us once with `fastlane`.
Sure we can, I'll add it to the dependencies module.

christopherfujino (approved):
LGTM

PR #215, opened 2020-09-15
Add android licenses for builds in the engine.

godofredoc:
Builds in the engine have not migrated to use the licenses from cipd.


godofredoc:
Landing as TBR as it will fix the engine aot builder.

config/engine_config.star
```diff
@@ -107,6 +107,9 @@ def engine_properties(
         "build_android_jit_release": build_android_jit_release,
         "gcs_goldens_bucket": gcs_goldens_bucket,
     }
+    if any([build_android_debug or build_android_aot or build_android_vulkan or build_android_jit_release]):
+        properties["android_sdk_license"] = "\n24333f8a63b6825ea9c5514f83c2829b004d1fee"
+        properties["android_sdk_preview_license"] = "\n84831b9409646a918e30573bab4c9c91346d8abd"
```
godofredoc, line 112: can we use the cipd package instead?  https://cs.opensource.google/flutter/recipes/+/master:recipe_modules/android_sdk/api.py

godofredoc: E.g. you can add a dependencies property to the builder config and then use 

```
  env = {}
  env_prefixes = {}
  deps = api.properties.get('dependencies', [])
  api.flutter_deps.required_deps(env, env_prefixes, deps)
  with api.context(env=env, env_prefixes=env_prefixes):
       # do something here

PR #219, opened 2020-09-17
Split hostonly devicelab linux tests.

godofredoc:
This is to move the devicelab tests to its own recipe.

Bug:
  https://github.com/flutter/flutter/issues/63242


keyonghan (approved):
LGTM

PR #222, opened 2020-09-18
Use shorter names to fit more tests in the view area.

godofredoc:
We were using very long names as short names using a lot of space in the
dashboard. This will allow us to see more tests without having to
scroll.


keyonghan (approved):
LGTM

PR #223, opened 2020-09-22
[web] add new builders to run the new recipe

nturgut:
Adding new try and prod builder to use the new recipe for running framework test with web engine. (Recipe: https://flutter.googlesource.com/recipes/+/refs/heads/master/recipes/engine/web_engine_framework.py)

Note: I got an error when running lucicfg:

```
luci.builder("try/Linux Web Framework tests") refers to undefined luci.executable("engine/web_engine_framework")
```

please let me know which steps I should take to add the new recipe to recupe_bundles CIPD package (or is it an automated process)

Fixes: https://github.com/flutter/flutter/issues/65052

### DIFF
--- a/config/engine_config.star
+++ b/config/engine_config.star
@@ -221,6 +221,7 @@ def engine_prod_config(platform_args, branch, version, ref, fuchsia_ctl_version)
         priority = 30 if branch == "master" else 25,
         **platform_args["linux"]
     )
+
     # Defines engine Linux builders
     common.linux_prod_builder(
         name = builder_name("Linux%s Host Engine|host", branch),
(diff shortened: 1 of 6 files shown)


config/engine_config.star
```diff
@@ -74,6 +74,9 @@ def engine_properties(
         build_android_jit_release = False,
         no_bitcode = False,
         no_lto = False,
+        framework = False,
```
godofredoc, line 77: you need to run:

```
lucicfg fmt .
lucicfg lint .
./config/main.star
```

For this to work you'll need depot tools in your path.

nturgut: Yes we also use depot_tools for building the engine so it was in the path.

I did the formating/linting but I'm still not able pass the undefined executable errors. Is there an example on how to use https://github.com/flutter/infra/blob/master/config/lib/recipes.star#L20?

nturgut: done. thanks!

godofredoc (approved):

PR #226, opened 2020-09-30
Add builders to led-recipe-tests.

godofredoc:
The flutter_drone and and adhoc portions of the recipes are not being
exercised on the led recipes testing. We are adding some of the lighter
builders to the led recipe tests.


keyonghan (approved):
LGTM

PR #229, opened 2020-10-01
Add deploy_gallery linux try builder config

keyonghan:
This PR adds linux try config for `deploy_gallery` test.

Related issue: https://github.com/flutter/flutter/issues/66707

### DIFF
--- a/config/framework_config.star
+++ b/config/framework_config.star
@@ -609,6 +609,22 @@ def framework_try_config():
             swarming.cache(name = "pub_cache", path = ".pub_cache"),
         ],
     )
+    common.linux_try_builder(
+        name = "Linux deploy_gallery|dg",
+        recipe = "flutter/flutter",
+        repo = repos.FLUTTER,
+        list_view_name = list_view_name,
+        properties = {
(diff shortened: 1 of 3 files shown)


config/framework_config.star
```diff
@@ -609,6 +609,22 @@ def framework_try_config():
             swarming.cache(name = "pub_cache", path = ".pub_cache"),
         ],
     )
+    common.linux_try_builder(
```
godofredoc, line 612: Can we also added to prod?

godofredoc (approved):

config/framework_config.star
keyonghan: Created this for testing. Once the recipe is validated, I will create a prod config.

godofredoc: You don't need to create the config to test the recipe you can do:

```led get-builder "luci.flutter.try: Linux docs" > properties.json```

Modify the properties section with the dependencies and secrets required, then do:

```cat properties.json | led edit-recipe-bundle | led launch```

keyonghan: This is handy! Will try it out.

PR #230, opened 2020-10-02
Fix codesigned test.

godofredoc:
The test was supposed to be running on mac but it was added as linux
test accidentally.

Bug:
  https://github.com/flutter/flutter/issues/67024


christopherfujino (approved):
LGTM

PR #237, opened 2020-10-13
Add flutter docs properties for firebase upload.

godofredoc:
These are the dependencies and properties required to upload logs to
firebase.

Bug:
  https://github.com/flutter/flutter/issues/60646


config/framework_config.star
```diff
@@ -242,6 +242,8 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
         properties = {
             "validation": "docs",
             "validation_name": "Docs",
+            "dependencies": [{"dependency": "dashing"}, {"dependency": "firebase"}],
+            "firebase_project": "master-docs-flutter-dev" if branch == "master" else "docs-flutter-dev",
```
keyonghan, line 246: Do we need firebase deploy for try builder?

keyonghan (approved):
LGTM with nit.

config/framework_config.star
godofredoc: No, try builders will only run the test without generating the docset and/or uploading.

PR #238, opened 2020-10-14
Remove docs from led-tests.

godofredoc:
This test relies on many environment variables that are not set
correctly on led-tests making them fail.

Bug:
  https://github.com/flutter/flutter/issues/60646


godofredoc:
Landing as TBR to complete the docs migration.

PR #239, opened 2020-10-14
Add release ref

godofredoc:
This is for the recipes to calculate the correct release ref for
deciding when to deploy changes or not.


godofredoc:
Landing as TBR to unblock uploading docs.

christopherfujino:
Sorry, I missed this PR. From reading the source of `//recipes/recipes/recipes.py` it looks like this gets populated in the environment as `LUCI_BRANCH`?

PR #240, opened 2020-10-14
Remove AndroidX test cases and fast start test case

jonahwilliams:
Infra config change for https://github.com/flutter/flutter/pull/68110


jonahwilliams:
These tests have dependencies on firebase libraries that are massive and need to be downloaded on every run. The AndroidX migration is 2+ years old at this point, so the value has diminished significantly.

godofredoc (approved):
Awesome! I like that we're getting some capacity back. Just one sanity check we need to remove these tests from *try.json and *prod.json before landing this PR.

PR #242, opened 2020-10-15
Increase execution timeout for windows to 1 and a half hours.

godofredoc:
Bug:
  https://bugs.fuchsia.dev/p/fuchsia/issues/detail?id=62070


cbracken (approved):
This seems reasonable as a short term measure, but if we're hitting builds this long once the goma work is resolved, we should look at sharding our builds further.

stuartmorgan (approved):
LGTM as a temporary measure.

PR #246, opened 2020-10-20
Add configurations to run bld_tests on Windows-Server bots.

godofredoc:
Windows server bots already contains all the dependencies required to
build flutter desktop.

Bug:
  https://github.com/flutter/flutter/pull/68657


config/framework_config.star
```diff
@@ -322,6 +322,7 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
             "shard": "build_tests",
             "subshards": ["0", "1_last"],
             "dependencies": [{"dependency": "android_sdk"}, {"dependency": "chrome_and_driver"}, {"dependency": "open_jdk"}, {"dependency": "goldctl"}],
+            "drone_dimensions": ["os=Windows-Server"],
```
keyonghan, line 325: I guess there will be a recipe side change to include this?

keyonghan (approved):
LGTM

config/framework_config.star
godofredoc: That is correct, the new property won't have any side effects but the change in recipes needs to remove the cores dimension to accept the new machines which are 32 cores instead of 8.

PR #247, opened 2020-10-20
Update fuchsia_ctl timeout.

godofredoc:
The fuchsia builds are taking more than one hour when starting from a
cold cache. Updating the timeout to 1 hour and a half to unblock led
tests.


keyonghan (approved):
LGTM

PR #250, opened 2020-10-23
Add dependencies to Linux build_tests.

godofredoc:
This is to support building application for Flutter desktop.

Bug:
  https://github.com/flutter/flutter/issues/58624


keyonghan (approved):
LGTM

PR #260, opened 2020-11-07
Update framework builders to use Windows-Server bot.

godofredoc:
This is to use the same images for all the windows bots and make
efficient use of capacity.

Bug: https://github.com/flutter/flutter/issues/68056


godofredoc:
Landing as TBR to minimize the impact during working hours.

PR #263, opened 2020-11-12
Remove x64 dimension.

godofredoc:
Adding this dimension by default is not needed and it is preventing the
devicelab staging tasks from finding the correct bots.

Bug:
   https://github.com/flutter/flutter/issues/70258


keyonghan (approved):
LGTM

PR #266, opened 2020-11-13
Do not block recipes on web_tool_tests.

godofredoc:
These tests are still failing and currently they are blocking landing
recipes changes and recipe dep rolls.


keyonghan (approved):
LGTM

PR #276, opened 2020-11-18
Upload DeviceLab metrics on master

CaseyHillers:
# Issues

https://github.com/flutter/flutter/issues/66191


config/devicelab_config.star
```diff
@@ -216,6 +216,7 @@ def devicelab_prod_config(branch, version, ref):
         properties = {
             "dependencies": [{"dependency": "android_sdk"}, {"dependency": "chrome_and_driver"}],
             "task_name": "web_benchmarks_html",
+	    "upload_metric": branch == "master",
```
christopherfujino, line 219: How about:

```
UPLOAD_METRICS_CHANNELS = ("master",)
```

and then:

```
  "upload_metric": True if branch in UPLOAD_METRICS_CHANNELS else False
```

That way it's more clear what's going on, and then adding release branches is a simple update. WDYT?

CaseyHillers: Thanks, done!

christopherfujino (approved):
LGTM, thanks!

PR #277, opened 2020-11-18
Increase staging builders expiration_timeout to be 12 hours

keyonghan:
We are running all mac/linux devicelab tests in staging pool, but with 2 bots for each platform. The default 6 hours make it difficult to see whether all tests pass.

This PR increase the expiration_timeout to be 12 hours to have a better visibility.

### DIFF
--- a/config/lib/timeout.star
+++ b/config/lib/timeout.star
@@ -9,5 +9,7 @@ timeout = struct(
     MEDIUM = 60 * time.minute,
     LONG = 90 * time.minute,
     XL = 180 * time.minute,
+    # EXPIRATION is used to set builder expiration_timeout, which controls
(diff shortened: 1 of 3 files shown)


config/lib/timeout.star
```diff
@@ -9,4 +9,5 @@ timeout = struct(
     MEDIUM = 60 * time.minute,
     LONG = 90 * time.minute,
     XL = 180 * time.minute,
+    LONG_EXPIRATION = 12 * 60 * time.minute,
```
godofredoc, line 12: Can we add a comment explaining where these values are expected to be used?

keyonghan: Added.

godofredoc (approved):

PR #280, opened 2020-11-25
Add a test shard for flutter/plugins

amirh:
Run the flutter/plugins tests (configured in https://github.com/flutter/flutter/blob/2ceb371d214cfe9122c2aeec7625863b21a1e41c/dev/bots/test.dart#L148) on post-submit.

Note that I'm not at all familiar with this config format, I'm pretty much blindly following https://g3doc.corp.google.com/company/teams/flutter/luci/builders.md?cl=343188502. Specifically I was unsure about which dependencies to specify, this test needs Flutter, git, and pub on the machine, I'm assuming that's part of the default environment?

This is part of: https://github.com/flutter/flutter/issues/69309


amirh:
I don't really know what I'm doing here (the in-progress tutorial doesn't have much details on the steps for adding a configuration here), I got the following message on CI:

```
echo $SERVICE_ACCOUNT > service_account.json
lucicfg validate ./config/main.star -service-account-json service_account.json
[P58 05:03:21.074 generate.go:94 I] Generating configs...
Error: the following files need to be regenerated: luci/cr-buildbucket.cfg, luci/luci-milo.cfg, luci/luci-notify.cfg, luci/luci-scheduler.cfg.
  Run `lucicfg generate "./config/main.star"` to update them.
```

So I updated the PR with the results of `lucicfg generate "./config/main.star"`, let me know if that's not what I'm supposed to do.

christopherfujino:
> So I updated the PR with the results of `lucicfg generate "./config/main.star"`, let me know if that's not what I'm supposed to do.

This should work. I added a comment to the WIP doc CL to execute `$ ./config/main.star`, which will essentially do the same.

config/framework_config.star
```diff
@@ -344,6 +344,22 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
             swarming.cache(name = "android_sdk", path = "android29"),
         ],
     )
+    common.linux_prod_builder(
+        name = "Linux%s flutter_plugins|fltplgns" % ("" if branch == "master" else " " + branch),
+        recipe = drone_recipe_name,
```
christopherfujino, line 349: i think this should be `new_recipe_name`.

config/framework_config.star
```diff
@@ -344,6 +344,22 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
...
+        console_view_name = console_view_name,
+        triggered_by = [trigger_name],
+        triggering_policy = triggering_policy,
+        properties = {
+            "shard": "flutter_plugins",
+            "subshard": "analyze",
```
christopherfujino, line 355: You can delete this key. As per above, this will not be a sub-shard of another builder.

config/framework_config.star
christopherfujino: Ok, never mind. I'm guessing you want to schedule sub-builds because you plan to add more shards in the future?

config/framework_config.star
```diff
@@ -344,6 +344,22 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
...
+            "dependencies": [],
+        },
+        caches = [
+            swarming.cache(name = "pub_cache", path = ".pub_cache"),
+            swarming.cache(name = "android_sdk", path = "android29"),
```
christopherfujino, line 360: I think either your shard requires andrroid sdk, in which case you should add it as a dependency, or it doesn't, in which case you should remove this cache directory.

config/framework_config.star
```diff
@@ -344,6 +344,22 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
             swarming.cache(name = "android_sdk", path = "android29"),
         ],
     )
+    common.linux_prod_builder(
```
christopherfujino, line 347: i think you also want a try builder.

config/framework_config.star
amirh: I actually followed the tutorial and copy-pasted the configuration for a test with a single sub-shard.
I do see this getting more sub-shards eventually, but I had thought that when that happens I'll follow the tutorial and copy-paste a sample with multiple shards 😄 

If I understand the tutorial correctly, it says that the "drone" recipes are used for single shards?

config/framework_config.star
amirh: done

config/framework_config.star
amirh: done

amirh (commented):
Thanks, PTAL

config/framework_config.star
godofredoc: using the drone recipe works the best for a single shard.

godofredoc (approved):

PR #281, opened 2020-11-30
adding web drones which will be called from web engine code

nturgut:
This is for adding web engine drone builders.

These will be used from the web engine recipe, for sharding (which will eliminate the flakiness happening now on linux web engine) The related design doc: go/web-engine-recipe-drones

related: https://github.com/flutter/flutter/issues/71576

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,5 @@
 {
   "files.associations": {
-    "*.star": "python"
+    "*.star": "starlark"
   }
 }
--- a/config/engine_config.star
+++ b/config/engine_config.star
@@ -61,6 +61,7 @@ def engine_recipes(version):
         "femu_test",
         "engine/engine_metrics",
         "engine/scenarios",
+        "engine/web_engine_drone",
         "engine/web_engine_framework",
     ]
     for name in recipe_list:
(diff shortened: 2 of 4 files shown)


config/engine_config.star
```diff
@@ -235,6 +236,15 @@ def engine_prod_config(platform_args, branch, version, ref, fuchsia_ctl_version)
         priority = 30 if branch == "master" else 25,
         **platform_args["windows"]
     )
+    common.linux_prod_builder(
+        name = builder_name("Linux%s Web Drone|webdrn", branch),
```
yjbanov, line 240: I'm not well-versed in LUCI terminology, but I thought "drones" were subtasks of a "builder". If so, then calling the builder itself "Drone" (and "webdrn") may be confusing. Since this builder will contain all web shards, can we just call it "Web"?

yjbanov (approved):
LGTM if LGT @christopherfujino

config/engine_config.star
nturgut: I used the word drone since the other ones in the framework side are called. SDK drones.

nturgut: You can search for Drone in this file: https://github.com/flutter/infra/blob/509d65c6172f0434740b5423f182a17411aaa558/config/framework_config.star

yjbanov: SGTM then.

christopherfujino (approved):
LGTM

.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
   "files.associations": {
-    "*.star": "python"
+    "*.star": "starlark"
```
christopherfujino, line 3: was this intentional? I don't use vscode, so i have no preference, just curious

nturgut: Yes, the highlights and all other IDE help is better when we set for starlark :)

PR #282, opened 2020-12-03
Use ios for device_os instead of specific versions.

godofredoc:
This is to use the two devices we currently have in staging.


godofredoc:
Landing as TBR to start using the two staging devices.

PR #283, opened 2020-12-03
Increase timeouts for linux web engine builders to 90mins.

godofredoc:
These builders are very close to the time limit causing timeouts from time to time.


config/engine_config.star
```diff
@@ -214,6 +214,7 @@ def engine_prod_config(platform_args, branch, version, ref, fuchsia_ctl_version)
         triggered_by = [trigger_name],
         triggering_policy = triggering_policy,
         priority = 30 if branch == "master" else 25,
+        execution_timeout = timeout.LONG,
```
yjbanov, line 217: This is probably going to go away entirely when we switch to shards, but just in case it might be worth leaving a TODO.

yjbanov (approved):
lgtm

config/engine_config.star
yjbanov: I filed https://github.com/flutter/flutter/issues/71692

cbracken (approved):
Thanks!

nturgut:
@godofredoc I merged sharding recipe change. Looks like it's working without any issues for a few days now. Can we revert this PR?

godofredoc:
> @godofredoc I merged sharding recipe change. Looks like it's working without any issues for a few days now. Can we revert this PR?

Sound good to me, thanks!

PR #286, opened 2020-12-09
Update Xcode version on plugin_lint_mac builder

jmagman:
Add custom configurations for plugin_lint_mac to update Xcode version from 11a420a to 11e708.

Will hopefully allow this test to build macOS projects (there's something weird going on with the swift version shipped in that toolchain).

https://github.com/flutter/flutter/pull/71953


config/generated/flutter/luci/cr-buildbucket.cfg
```diff
@@ -9046,27 +9046,22 @@ buckets {
...
         cipd_version: "refs/heads/master"
-        properties_j: "$depot_tools/osx_sdk:{\"sdk_version\":\"11a420a\"}"
+        properties_j: "$depot_tools/osx_sdk:{\"sdk_version\":\"11E708\"}"
         properties_j: "$fuchsia/goma:{\"server\":\"rbe-prod1.endpoints.fuchsia-infra-goma-prod.cloud.goog\"}"
         properties_j: "$kitchen:{\"emulate_gce\":true}"
         properties_j: "$recipe_engine/isolated:{\"server\":\"https://isolateserver.appspot.com\"}"
         properties_j: "$recipe_engine/swarming:{\"server\":\"https://chromium-swarm.appspot.com\"}"
         properties_j: "clobber:false"
-        properties_j: "dependencies:[{\"dependency\":\"android_sdk\"},{\"dependency\":\"open_jdk\"},{\"dependency\":\"xcode\"},{\"dependency\":\"gems\"}]"
+        properties_j: "dependencies:[{\"dependency\":\"xcode\"},{\"dependency\":\"gems\"}]"
```
jmagman, line 9058: I don't think this test needs android_sdk or open_jdk.

godofredoc (approved):

config/framework_config.star
```diff
@@ -511,6 +511,22 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
             swarming.cache(name = "android_sdk", path = "android29"),
         ],
     )
+    common.mac_prod_builder(
```
jmagman, line 514: Nothing else in this file has a `task_name` set, does it matter that dropped off?

config/generated/flutter/luci/cr-buildbucket.cfg
```diff
@@ -9046,27 +9046,22 @@ buckets {
       dimensions: "os:Mac-10.15"
       dimensions: "pool:luci.flutter.prod"
       recipe {
-        name: "devicelab/devicelab_drone_1_24_0"
+        name: "flutter/flutter_drone_1_24_0"
```
jmagman, line 9049: Does it matter this changed from `devicelab/devicelab_drone` to `flutter/flutter_drone`?

godofredoc: probably yes, the recipes are different if this is actually a devicelab test using the devicelab runner it should be on the devicelab_config.

config/framework_config.star
godofredoc: Taking a closer look this won't work. Flutter sdk configurations use the sharded test runner and it knows what to run based on shard and subshard. If this uses the devicelab runner then it needs the task name.

config/generated/flutter/luci/cr-buildbucket.cfg
jmagman: I'm confused, because it's not a devicelab test, it doesn't need an attached device.  It's currently running on every PR and in post-submit on LUCI builders.
https://ci.chromium.org/p/flutter/builders/prod/Mac%20plugin_lint_mac

config/framework_config.star
jmagman: But this one isn't sharded.

jmagman: Oh, are you talking about these shards?  This one isn't using that runner.

https://github.com/flutter/flutter/blob/4300226a50670cff49693dfa26a2d6844238718d/dev/bots/test.dart#L122-L134

jmagman: The task name here is "mac_plugin_lint_mac " which doesn't match the "plugin_lint_mac" once which was removed. https://github.com/flutter/flutter/blob/eed9835a165a6db563f33fc73a61cf2621519757/dev/try_builders.json#L282

godofredoc: The initial configurations(the ones being deleted) are in the correct place, we only need to add:

```
"$depot_tools/osx_sdk": {
                "sdk_version": "11E708",
            },
```
To its properties

config/generated/flutter/luci/cr-buildbucket.cfg
christopherfujino: Yeah, but it's running the devicelab recipe not because it runs in the devicelab, but because it uses the devicelab test runner (`//dev/devicelab/bin/run.dart`), whereas the framework recipe uses `//dev/bots/test.dart`.

jmagman: Yeah I got confused about WHICH runner:
> Oh, are you talking about these shards?  This one isn't using that runner.
> https://github.com/flutter/flutter/blob/4300226a50670cff49693dfa26a2d6844238718d/dev/bots/test.dart#L122-L134

https://github.com/flutter/infra/pull/286#discussion_r538999625

config/framework_config.star
jmagman: 🤦‍♀️  of course.  Thanks for pointing it out!

PR #287, opened 2020-12-09
Move ios_content_validation_test from staging devicelab to devicelab (host builder tests)

jmagman:
`ios_content_validation_test` doesn't need to run against a real device and can be moved from the devicelab to a LUCI builder like the other previous `hostonly_devicelab_tests` tests.

PR #262, opened 2020-11-11
Initial configuration for mac devicelab staging.

godofredoc:
This PR enables a mac devicelab test running on luci and using the
staging environment.

Bug:
  https://github.com/flutter/flutter/issues/70258

PR #288, opened 2020-12-10
Re-add ios_signing dependency to ios_content_validation_test

jmagman:
https://github.com/flutter/infra/pull/287 moved ios_content_validation_test from devicelab to luci builders.  It failed on a try build with `No valid code signing certificates were found`.  Re-add `ios_signing` dependency removed in that PR.

https://github.com/flutter/infra/pull/287#discussion_r539705796


jmagman:
@keyonghan I know this dependency was introduced for the devicelab staging in #262.  Will it also work in this config?

keyonghan:
> @keyonghan I know this dependency was introduced for the devicelab staging in #262. Will it also work in this config?

Does this test need a phone device? We do not have any bot with a device available in the prod pool at this moment.

jmagman:
> Does this test need a phone device? We do not have any bot with a device available in the prod pool at this moment.

No, it doesn't need a device, that's why I moved it in https://github.com/flutter/infra/pull/287.  It's archiving the app, which requires a provisioning profile.

config/devicelab_config.star
```diff
@@ -541,6 +541,9 @@ def devicelab_prod_config(branch, version, ref):
                 {
                     "dependency": "gems",
                 },
+                {
```
godofredoc, line 544: This dependency is only available in staging and not available yet for prod. The reason is signing will be only supported on devicelab_machines where we control the provisioning.

jmagman: Filed https://github.com/flutter/flutter/issues/72078

jmagman:
Closing, looks like this test can't be moved from `devicelab_machines`.

PR #289, opened 2020-12-10
Remove device_os dimension from linux test beds to unblock migration.

godofredoc:
The library to collect phone dimensions is not compatible with adb
daemon and we will need to collect device dimensions using
device_doctor.


CaseyHillers (approved):
LGTM

keyonghan (approved):
LGTM

PR #290, opened 2020-12-10
Revert "Move ios_content_validation_test from staging devicelab to devicelab (host builder tests)"

jmagman:
Reverts flutter/infra#287

This test can't be moved from `devicelab_machines` because iOS provisioning profiles aren't available on prod.  Filed https://github.com/flutter/flutter/issues/72078


keyonghan (approved):
LGTM

PR #291, opened 2020-12-10
Add macos_content_validation_test

jmagman:
Create new `macos_content_validation_test`.  Will be similar to ios_content_validation_test but doesn't need to live in the devicelab, and doesn't need provisioning profiles.


keyonghan (approved):
LGTM

PR #292, opened 2020-12-10
Add open_jdk dependency to luci devicelab linux tasks.

godofredoc:
Most of the tasks require this dependency to build.


CaseyHillers (approved):
LGTM

PR #295, opened 2020-12-11
Get the linux web engine timeout back to the default 1 hour.

godofredoc:
This is actually a revert of https://github.com/flutter/infra/pull/283/files but required a new pr because it didn't revert cleanly.


nturgut (approved):
LGTM. Thanks!

PR #301, opened 2020-12-15
Transtionally support both flutter and recipe_engine osx modules.

godofredoc:
These is to make three steps transtion to avoid time down.


jmagman (approved):
LGTM

PR #302, opened 2020-12-15
Move devicelab staging to use flutter osx module.

godofredoc:


godofredoc:
Landing as TBR to unblock staging bots.

PR #304, opened 2020-12-15
Use xcode 11.6 from flutter packages for the sdk.

godofredoc:
This is in preparation to use xcode internal packages in the engine, sdk
and devicelab.


jmagman (approved):
LGTM

PR #305, opened 2020-12-15
Add devicelab linux tasks to prod

keyonghan:
Devicelab linux tasks have been running successfully in staging for couple of weeks. This PR enables these tasks in prod.

Related issue: https://github.com/flutter/flutter/issues/71615


config/devicelab_config.star
```diff
@@ -222,14 +225,90 @@ def devicelab_prod_config(branch, version, ref):
...
+                        "dependency": "open_jdk",
+                    },
+                ],
+                "task_name": task,
+                "upload_metrics": branch in UPLOAD_METRICS_CHANNELS,
+            },
+            pool = "luci.flutter.prod",
+            os = "Android",
+            dimensions = {"device_os": "N"},
+            expiration_timeout = timeout.LONG_EXPIRATION,
```
CaseyHillers, line 302: Is this necessary? Shouldn't devicelab tests take less than 15 minutes?

CaseyHillers (approved):
LGTM w/ comment on timeout

yusufm (approved):
lgtm

config/devicelab_config.star
keyonghan: Based on observation from devicelab task queue time, the 90th% is around 3-4 hours. However, from LUCI side, with batch run enabled, the queue time should improve.
But I would prefer to keep it as it is now to avoid unexpected kill. Will adjust the timeout when everything is stable.

godofredoc: With the new scheduling it is expected that tasks stay a max of 15 mins in the queue and align with the current SLOs. How long it is taking for the current tasks to start the execution?

CaseyHillers: SGTM to leave this, but we should have a tracking issue to bring it back in line with the SLO.

keyonghan: Checked the queue time of staging (since we enabled 6 bots), the 90th% is ~116 min.

godofredoc: Are we doing something wrong in the scheduling? we have a max batch size of 20 with two max concurrent builds I would not expect more than 15 mins queue time.

godofredoc (approved):
Let's land the change and open a bug to investigate the long queue time.

config/devicelab_config.star
keyonghan: Scheduling seems to work as expected.
The long queue time (90th%) should due to cases where new commits keep jumping in, and the scheduler `randomly` (not sure if it is random, can we control this? like running those old tasks with higher priority) picks up tasks (may leave some tasks not picked up for a long time even the batch size is 20).

keyonghan:
> Let's land the change and open a bug to investigate the long queue time.

https://github.com/flutter/flutter/issues/72383

config/devicelab_config.star
godofredoc: The scheduler is deterministic and probably we got the meaning of the config properties wrong. E.g the expectation is that a max of two commits will be tested in parallel and that it will bundle up to 20 commits in a batch if there is more than a commit.

PR #307, opened 2020-12-15
Allow only 5 devicelab linux tasks initially

keyonghan:
This is a followup of https://github.com/flutter/infra/pull/305.


config/devicelab_config.star
```diff
@@ -231,47 +231,6 @@ def devicelab_prod_config(branch, version, ref):
         "android_defines_test",
         "android_obfuscate_test",
         "android_view_scroll_perf__timeline_summary",
-        "animated_placeholder_perf__e2e_summary",
```
godofredoc, line 234: If they are already here let's keep them and delete most of the tasks from staging instead. As long as we don't add them to the prod.json file and remove them from the devicelab yaml files it does not affect if they don't pass or are slow.

PR #311, opened 2020-12-16
Change `max_concurrent_invocations` to 1 for luci devicelab builders

keyonghan:
There are five bots running 45 LUCI devicelab builders.

Current setting `max_concurrent_invocations=3` makes the queue time super high:
Screen Shot 2020-12-16 at 12 37 24 PM

Like in the above picture, each builder has three queued builds. This will make the tot builds queuing time even higher if we have continuous more new commits joining. (after finish running a build, it may choose to run an older queued build).

Updating `max_concurrent_invocations` to 1 should have the scheduler always choose newer build to run.

From [quote](https://chromium.googlesource.com/infra/luci/luci-go/+/refs/heads/master/lucicfg/doc/README.md#scheduler.policy-args)
> max_concurrent_invocations: limit on a number of builds running at the same time. If the number of currently running builds launched through LUCI Scheduler is more than or equal to this setting, LUCI Scheduler will keep queuing up triggering requests, waiting for some running build to finish before starting a new one. Default is 1.

### DIFF
--- a/config/devicelab_config.star
+++ b/config/devicelab_config.star
@@ -78,7 +78,7 @@ def devicelab_prod_config(branch, version, ref):
     if branch == "master":
         triggering_policy = scheduler.greedy_batching(
             max_batch_size = 3,
-            max_concurrent_invocations = 3,
+            max_concurrent_invocations = 1,
         )
     else:
         triggering_policy = scheduler.greedy_batching(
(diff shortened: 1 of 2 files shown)


yusufm (approved):
should we increase the max_batch_size?

lgtm

keyonghan:
> should we increase the max_batch_size?
> 
> lgtm

Maybe not. We want to run builds at least every three commits.

PR #314, opened 2020-12-21
Remove swift dependency as it is now included with xcode.

godofredoc:
Now that flutter has its own xcode packages with swift already included
we don't need this complexity anymore.

Bug:
  https://github.com/flutter/flutter/issues/72301
  https://github.com/flutter/flutter/issues/72280


keyonghan (approved):
LGTM

CaseyHillers (approved):
LGTM!

PR #315, opened 2020-12-22
Run framework build tests in recipes presubmit.

godofredoc:
These tests have swift dependencies that need to be tested on presubmit
for changes to the recipes.

Bug:
  https://github.com/flutter/flutter/issues/72823


keyonghan (approved):
LGTM

PR #316, opened 2020-12-22
Start optimizing caches for devicelab.

godofredoc:
Devicelab builds are taking an extra 10mins downloading dependencies.
This is the first PR for the caches optimization.

Bug: https://github.com/flutter/flutter/issues/72740


keyonghan (approved):
LGTM

config/lib/common.star
```diff
@@ -381,16 +368,10 @@ def _linux_builder(
         category = "Linux",
         os = None,
         **kwargs):
-    linux_caches = [
```
CaseyHillers, line 384: Why are we removing the caches from all builders, but only adding to devicelab staging?

godofredoc: Because we will be optimizing the caches in a per repo/builder basis and these global configurations conflict with the per repo/builder configs. After landing this PR will send a few more for the different repos.

CaseyHillers: We should add this to the prod freeze changes sheet as some issues might not surface until we have more traffic in the new year.

CaseyHillers (approved):

config/lib/common.star
godofredoc: done.

PR #317, opened 2020-12-23
Add try builder config for cocoon device doctor

keyonghan:
Cocoon device doctor recipe to build/upload device_doctor package to cipd is ready. This PR adds pre-submit builder configs.

These builders should be triggered whenever any updates from `cocoon/device_doctor`.

### DIFF
--- a/config/cocoon_config.star
+++ b/config/cocoon_config.star
@@ -18,25 +18,18 @@ def _setup():
             "caches": [swarming.cache(name = "dart_pub_cache", path = ".pub-cache")],
         },
     }
-    cocoon_define_recipes()
-    device_doctor_recipes()
+    cocoon_recipes()
     cocoon_try_config(platform_args)
 
-def cocoon_define_recipes():
-    # Defines recipes
-    luci.recipe(
-        name = "cocoon",
-        cipd_package = "flutter/recipe_bundles/flutter.googlesource.com/recipes",
-        cipd_version = "refs/heads/master",
-    )
+recipe_names = ["cocoon/cocoon", "cocoon/device_doctor"]
 
-def device_doctor_recipes():
-    # Defines recipes
-    luci.recipe(
-        name = "cocoon/device_doctor",
-        cipd_package = "flutter/recipe_bundles/flutter.googlesource.com/recipes",
-        cipd_version = "refs/heads/master",
-    )
+def cocoon_recipes():
+    for recipe in recipe_names:
+        luci.recipe(
+            name = recipe,
+            cipd_package = "flutter/recipe_bundles/flutter.googlesource.com/recipes",
+            cipd_version = "refs/heads/master",
+        )
 
 def cocoon_try_config(platform_args):
     """Creates try cocoon configurations.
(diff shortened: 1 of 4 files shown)


config/cocoon_config.star
```diff
@@ -29,15 +30,29 @@ def cocoon_define_recipes():
...
     )
 
+def device_doctor_recipes():
+    # Defines recipes
+    luci.recipe(
+        name = "cocoon/device_doctor",
+        cipd_package = "flutter/recipe_bundles/flutter.googlesource.com/recipes",
+        cipd_version = "refs/heads/master",
+    )
+
```
godofredoc, line 40: Can we consolidate the two methods in a generic one? eg.
recipe_names = [ "cocoon", "cocoon/device_doctor"] 
for recipe in recipe_names:
   luci.recipe(
         name = recipe ...

config/cocoon_config.star
```diff
@@ -29,15 +30,29 @@ def cocoon_define_recipes():
         cipd_version = "refs/heads/master",
     )
 
+def device_doctor_recipes():
```
godofredoc, line 33: can we also move the cocoon recipe to cocoon/cocoon?

config/cocoon_config.star
```diff
@@ -46,5 +61,30 @@ def cocoon_try_config(platform_args):
         add_cq = True,
         **platform_args["linux"]
     )
+    common.linux_try_builder(
+        name = "Linux device_doctor|device_doctor",
+        recipe = "cocoon/device_doctor",
+        list_view_name = list_view_name,
+        repo = repos.COCOON,
+        add_cq = True,
```
godofredoc, line 69: do we need the three builders blocking recipes changes? if not we can remove add_cq from some of them.

config/cocoon_config.star
```diff
@@ -46,5 +61,30 @@ def cocoon_try_config(platform_args):
...
+    )
+
+    # Defines cocoon mac try builders
+    common.mac_try_builder(
```
godofredoc, line 73: Do the test require phones? if yes we'll need to add additional dimensions.

config/cocoon_config.star
keyonghan: First recipe change: https://flutter-review.googlesource.com/c/recipes/+/9520
Also updated here to point to cocoon/cocoon.
Will land this PR when the above CL lands.

config/cocoon_config.star
keyonghan: Done.

config/cocoon_config.star
keyonghan: This is intended to make sure these three builders succeed when any changes from recipe side.

config/cocoon_config.star
keyonghan: Not now until we have capacity in try pool.

config/cocoon_config.star
godofredoc: I think the question is mac/linux use the same code path do we need to run both in led recipe tests?

godofredoc (approved):

config/cocoon_config.star
keyonghan: I think so. Though they are using the same code path, but they fetch different versions of dart, and then proceed to build.

CaseyHillers (approved):

PR #321, opened 2020-12-23
Update cache configurations.

godofredoc:
This also adds the option to cache different versions of xcode.

Bug: https://github.com/flutter/flutter/issues/72740


keyonghan (approved):
LGTM

PR #322, opened 2020-12-23
Add an execution timeout of 30 mins for devicelab tests.

godofredoc:
This is to prevent tasks running for hours when they fail to connect to
the services.


keyonghan (approved):
LGTM

PR #325, opened 2021-01-09
Update xcode version to fix swift issue of devicelab mac builders

keyonghan:
This is to fix: https://github.com/flutter/flutter/issues/73606, by updating the xcode version.


config/devicelab_config.star
```diff
@@ -636,7 +637,7 @@ def devicelab_prod_config(branch, version, ref):
                 "sdk_version": XCODE_VERSION,
             },
             "$flutter/osx_sdk": {
-                "sdk_version": XCODE_VERSION,
+                "sdk_version": NEW_XCODE_VERSION,
```
godofredoc, line 640: I thought we were just updating the three failing builders

jmagman: Can we make sure Swift is available in all the Xcode packages?  Swift is the default language for new Flutter projects, this is becoming a moving target.  I just had to switch my tools tests to use non-default Objective-C to work around this:
https://github.com/flutter/flutter/pull/73577/commits/2d14a6a23c7bd0f4c6a591cfa185f7cb6c36a3bc

keyonghan: we will test other builders and will update to use 12.3 version once all builders are verified to work.

godofredoc (approved):

config/devicelab_config.star
jmagman: I meant can we update the old version to also have Swift available?

godofredoc: We are at days of updating everything in the framework and devicelab to 12.3 is it worth spending some hours to update the old version to include swift?

jmagman: No I guess I don't care.  It would be nice to always have it available going forward though, without needing to have different builders know about which Xcode versions are safe to use.

godofredoc: Agree, that's one of the reasons we are now using our own bundles: https://g3doc.corp.google.com/company/teams/flutter/luci/xcode.md?cl=head

jmagman: Like this change doesn't include tool_tests and I have to know not to use Swift in that one even though that's what our users will use.  It would be better if I could assume Swift was available everywhere.  Which is fine if that's the expectation for every new version, we don't need to fix the old one.

jmagman: > Which is fine if that's the expectation for every new version, we don't need to fix the old one.

Sounds like that's the plan then (I sent my message before I saw yours).

keyonghan: Created https://github.com/flutter/flutter/issues/73623 to track.

Verified these three failed builders passed with updated xcode version:
https://chromium-swarm.appspot.com/task?id=50fd455a43bb9010
https://chromium-swarm.appspot.com/task?id=50fd57ee800a3e10
 https://chromium-swarm.appspot.com/task?id=50fd1a6c3f974c10

Merging this PR for now.

godofredoc: I think we should be explicit about what dependencies are needed for the tests so that the bundles include them on the future versions. Right now we have had to find what those dependencies are on a trial and error basis and we have been documenting them along with the process to create new packages.

We are still in the middle of making the xcode version consistent across all the repos once we get there we can safely assume, current and new versions will be consistent with test requirements.

PR #327, opened 2021-01-14
Add try builders for initial Linux and Windows desktop devicelab tests

stuartmorgan:
Try builders to run the new Linux and Windows devicelab tests. There is only one test, and only try (no prod) since this is to validate whether these configurations work for running desktop tests. If they are successful, they will be added to prod, and more tests will be rolled out.

Part of https://github.com/flutter/flutter/issues/54295


stuartmorgan:
This is my first time editing these files, so please give it very close scrutiny 🙂  I used the loop structure I saw in a couple of places rather than individual entries because I expect all the desktop tests would use the same configuration, so over time this should be much less verbose. If there's a reason not to do that, I can easily change it.

keyonghan (approved):
LGTM.

keyonghan (approved):
LGTM

config/devicelab_config.star
```diff
@@ -1493,4 +1520,32 @@ def devicelab_try_config():
...
+        "hot_mode_dev_cycle_windows_target__benchmark",
+    ]
+    for task in windows_desktop_tasks:
+        common.windows_try_builder(
+            name = "Windows %s|%s" % (task, short_name(task)),
+            recipe = drone_recipe_name,
+            repo = repos.FLUTTER,
+            list_view_name = list_view_name,
+            properties = {
+                "dependencies": [
```
godofredoc, line 1534: are we using clang and cmake to build windows applications? if we are using vs2019 it is already included in the image and we don't need to add them as additional dependencies.

stuartmorgan: This is copypasta. I remember thinking "I'll copy and paste this, change all the "linux"es to "windows", and remove the Linux dependencies". 2/3!

godofredoc (approved):

PR #328, opened 2021-01-14
Fix Windows desktop task name

stuartmorgan:
The task name abbreviates 'windows' to 'win', so the builder was trying to run the wrong task.


keyonghan (approved):
LGTM

PR #330, opened 2021-01-17
Change xcode bundle for staging.

godofredoc:
This also removes a test that do not exist anymore
(smoke_catalina_start_up_ios).


godofredoc:
Landing as TBR to unblock testing the new bundle.

PR #331, opened 2021-01-19
Enable hot reload devicelab test for Windows

stuartmorgan:
Since the try version passed, add it to prod so that it becomes part of
the postsubmit test suite.

Part of https://github.com/flutter/flutter/issues/54295


config/devicelab_config.star
```diff
@@ -869,6 +869,26 @@ def devicelab_prod_config(branch, version, ref):
         os = "Windows-Server",
     )
 
+    # Desktop Windows prod builders
+    # Currently Windows is only supported on master and dev.
+    if branch == "master" or branch == "dev":
```
stuartmorgan, line 874: Is this the right way to handle only having support on master and dev, or should that be controlled at a higher level? I'm unclear whether just having an entry in this file causes post-submit tests to run, or if it makes it possible to run them and actually running them is driven elsewhere. (I guessed the former due to the `trigger*` variables in the config).

keyonghan: I think this would be the way to go, but not 100% sure. @godofredoc to confirm?

stuartmorgan: @godofredoc Ping?

godofredoc: We are actually using release_ref instead of branch, eg.

```# Builders defined only for release refs
    if release_ref in (r"refs/heads/stable", r"refs/heads/beta", r"refs/heads/dev"):
```

stuartmorgan: Update, PTAL.

godofredoc (approved):

PR #332, opened 2021-01-21
Move mac_android tests to prod.

godofredoc:
Bug: https://github.com/flutter/flutter/issues/71960
Bug: https://github.com/flutter/flutter/issues/74059
Bug: https://github.com/flutter/flutter/issues/74142


yusufm (approved):

config/devicelab_staging_config.star
```diff
@@ -210,7 +210,7 @@ def devicelab_staging_prod_config():
 
     for task in mac_android_tasks:
         common.mac_prod_builder(
-            name = "Mac_android %s|%s" % (task, short_name(task)),
+            name = "Mac_android_staging %s|%s" % (task, short_name(task)),
```
keyonghan, line 213: I am wondering if we should use `Mac_android` prefix for both devicelab and devicelab_staging, since we already have the `console_view_name` to distinguish the consoles.

With same name, the builder names would be consistent, and also it will retain the historical builds (if we change to `Mac_android_staging`, I suppose the old builds with `Mac_android` will not be appended).

keyonghan (approved):
LGTM with nit.

config/devicelab_staging_config.star
godofredoc: The names need to be unique or we get an error about a duplicated task.

PR #337, opened 2021-01-29
Increase mac ios execution timeouts and add ios32 tests.

godofredoc:
Bug: https://github.com/flutter/flutter/issues/74920


godofredoc:
Adding ios32 to staging while we visit the lab to recover the mac/ios32 test bed in prod.

config/devicelab_staging_config.star
```diff
@@ -161,7 +161,43 @@ def devicelab_staging_prod_config():
...
+                        "dependency": "gems",
+                    },
+                    {
+                        "dependency": "ios_signing",
+                    },
+                ],
+                "task_name": task,
+            },
+            pool = "luci.flutter.staging",
+            os = "iOS-9.3.6",
```
jmagman, line 199: Is this the piece that tells it to only run on machines with this iOS version available?

godofredoc: Yes that is correct. Uploading an update to use iOS-14.3 for all the others.

godofredoc: Updated!

jmagman (approved):
LGTM, I guess when the iOS 14 devices update we'll have a window where they will be updating and `os = "iOS-14.3"` will be out of sync with some of them?  Sounds like logistics to be figured out later.

godofredoc:
> LGTM, I guess when the iOS 14 devices update we'll have a window where they will be updating and `os = "iOS-14.3"` will be out of sync with some of them? Sounds like logistics to be figured out later.

Yes, it will be added to the process for hardware/software updates documentation.

PR #344, opened 2021-02-03
Specify tasks to run on hosts without devices when not necessary.

yusufm:


config/engine_config.star
```diff
@@ -375,6 +376,7 @@ def engine_prod_config(platform_args, branch, version, ref, fuchsia_ctl_version)
         triggering_policy = triggering_policy,
         priority = 30 if branch == "master" else 25,
         execution_timeout = timeout.MEDIUM_LONG,
+        dimensions = {"device_type": "none"},
```
godofredoc, line 379: we are still missing the dimension on some of the mac_prod and mac_try builders.

yusufm:
Redoing this change in another PR.

PR #351, opened 2021-02-08
turned on flutter_engine_group_performance.dart

gaaclarke:
issue: https://github.com/flutter/flutter/issues/72026

depends on: https://github.com/flutter/flutter/pull/75374


jmagman (approved):
LGTM

gaaclarke:
What's the best way to babysit this to make sure it landed well, since a successful local run doesn't seem to guarantee it will run successfully on the bots?

gaaclarke:
I suspect results should be showing up at https://flutter-flutter-perf.skia.org/e/?queries=test%3Dflutter_engine_group_performance

keyonghan:
Before enabling it in flutter-dashboard, milo dashboard is where we check running status: https://ci.chromium.org/p/flutter/g/devicelab/console

PR #355, opened 2021-02-12
Use static xcode paths for mac ios 32 tests.

godofredoc:
Bug: https://github.com/flutter/flutter/issues/68322


config/devicelab_staging_config.star
```diff
@@ -184,6 +184,9 @@ def devicelab_staging_prod_config():
                 "$flutter/osx_sdk": {
                     "sdk_version": "12c33",  # 12.3
                 },
+                "$flutter/devicelab_osx_sdk": {
+                    "sdk_version": "12c33",  # 12.3
+                },
```
keyonghan, line 189: Do we need to keep these two versions?

godofredoc: Yes, at least during the transition we can delete flutter/osx_sdk once the tasks are using the new module.

CaseyHillers (approved):

keyonghan (approved):
LGTM

PR #358, opened 2021-02-19
Use xcode 12.3 consistently across framework and devicelab host tasks.

godofredoc:
This is in preparation to move all the devicelab mac/ios tasks to use
xcode 12.3.

Bug: https://github.com/flutter/flutter/issues/73392


config/devicelab_config.star
```diff
@@ -15,8 +15,7 @@ load("//lib/repos.star", "repos")
 load("//lib/timeout.star", "timeout")
 
 # Global xcode version for flutter/devicelab tests.
-XCODE_VERSION = "11e708"
-NEW_XCODE_VERSION = "12c5020f"
+NEW_XCODE_VERSION = "12c33"
```
jmagman, line 18: Can this be `XCODE_VERSION` now?

jmagman (approved):
LGTM with nit assuming all the tests pass.

config/devicelab_config.star
godofredoc: Renamed!

PR #365, opened 2021-02-24
Use ios14.4 to make all devices consistent.

godofredoc:
Updating all the devices to be used by luci mac/ios to 14.4 to make the configurations consistent.


godofredoc:
Landing as TBR to validate the setups while I'm in the lab

jmagman:
LGTM

PR #369, opened 2021-02-25
Remove failing 32-bit iOS benchmarks from staging

jmagman:
Removed in https://github.com/flutter/flutter/pull/76800.

See also https://github.com/flutter/flutter/issues/74920


jmagman:
I think native_ui_tests_ios32 and flutter_gallery__transition_perf_e2e_ios32 are safe to move to prod as unflaky.

godofredoc (approved):

PR #373, opened 2021-03-03
Add a config for plugin repo to trigger recipes when there are any tag changes.

cyanglaz:
Add a simple config to listen to tag changes in the plugin repo. 
This is to get ready for auto publish workflow for the plugin repo.

go/flutter-plugins-automatic-release


config/plugins_tagged_config.star
```diff
@@ -0,0 +1,57 @@
...
+def plugins_tagged_config_setup():
+
+    # Defines a list view for builders
+    list_view_name = "plugins-tagged"
+    luci.list_view(
+        name = list_view_name,
+        title = "Plugins tagged builders",
+    )
+
+    trigger_name = branch + "-gitiles-trigger-plugins-tagged"
```
christopherfujino: this can be:

`trigger_name = "gitiles-trigger-plugins-tagged"`

the var `branch` is `None` here.

config/main.star
```diff
@@ -28,6 +28,7 @@ load("//iostools_config.star", "iostools_config")
 load("//packages_config.star", "packages_config")
 load("//packaging_config.star", "packaging_config")
 load("//plugins_config.star", "plugins_config")
+load("//plugins_tagged_config.star", "plugins_tagged_config")
```
christopherfujino, line 31: You might consider including these in `plugins_config.star` and using the same `luci.list_view()`. that's a UX decision, but it would just mean less clicking if you want to compare both.

config/plugins_tagged_config.star
```diff
@@ -0,0 +1,71 @@
...
+    """Detailed builder configures."""
+
+    # Defines a list view for builders
+    list_view_name = "plugins-tagged"
+    luci.list_view(
+        name = list_view_name,
+        title = "Plugins tagged builders",
+    )
+
+    trigger_name = "-gitiles-trigger-plugins-tagged"
```
christopherfujino, line 37:
```suggestion
    trigger_name = "gitiles-trigger-plugins-tagged"
```

config/main.star
cyanglaz: Makes sense. I'll do that.

christopherfujino: we chatted offline, cyanglaz pointed out that this new builder is post-submit while the existing one is pre-submit, so it makes sense to have separate views.

config/generated/flutter/luci/cr-buildbucket.cfg
```diff
@@ -46917,6 +46917,31 @@ buckets {
       service_account: "flutter-prod-builder@chops-service-accounts.iam.gserviceaccount.com"
       task_template_canary_percentage {}
     }
+    builders {
```
cyanglaz, line 46920: These .cfg files are auto generated when running some setup command locally, do we check these in or not? @christopherfujino

config/main.star
cyanglaz: Done,  PTAL!

config/generated/flutter/luci/cr-buildbucket.cfg
christopherfujino: yep, everything under `config/generated` should be checked in. don't worry though, a CI pre-submit check will ensure that all the code is in sync.

config/plugins_config.star
```diff
@@ -57,4 +63,41 @@ def plugins_try_config(platform_args):
...
+        name = console_view_name,
+        repo = repos.PLUGINS,
+        refs = [ref],
+    )
+
+    publish_recipe_name = "plugins/plugins_publish"
+
+    # Defines builders
+    common.linux_prod_builder(
+        name = "Plugin Publish|publish",
```
christopherfujino, line 96: We ususally prefix the builder name with the OS, like `"Linux Plugin Publish|publish"`

cyanglaz: I originally did this because I saw other configs doing so. but the final name is ambiguous. It reads like "Linux Plugin", e.g. plugins for linux. If the prefixing builder name is an agreed pattern, then it's fine.

cyanglaz: actually "Linux publish plugins" is probably better

christopherfujino: Yeah, our builder names are unfortunate, but I think we should keep the convention because I am going to be writing tooling to parse it...

config/plugins_config.star
```diff
@@ -57,4 +63,41 @@ def plugins_try_config(platform_args):
...
+    # poll for any tags change
+    luci.gitiles_poller(
+        name = trigger_name,
+        bucket = "prod",
+        repo = repos.PLUGINS,
+        refs = [ref],
+    )
+
+    triggering_policy = scheduler.greedy_batching(
+        max_batch_size = 20,
```
christopherfujino, line 81: i forgot how the api for this works, but I don't think you want to batch commits together. That means if multiple commits land with tags within the github -> googlesource.com mirror window (15 minutes), only the latter commit will trigger a build. or does it not matter?

cyanglaz: This is a copy paste mistake, removed

christopherfujino (approved):
LGTM

PR #374, opened 2021-03-03
roll stable to branch config

christopherfujino:


keyonghan (approved):
Curious how these cfg file are generated based on json. do we have doc for that?

christopherfujino:
> Curious how these cfg file are generated based on json. do we have doc for that?

I can add more to the https://github.com/flutter/infra/blob/master/config/lib/release_branches/README.md

This is the secret sauce: https://github.com/flutter/infra/blob/master/config/lib/release_branches/release_branches.star

PR #375, opened 2021-03-04
Splits docs job into docs_test & docs_publish

christopherfujino:
Currently, we run the docs job on master + the three release candidate branches. if the branch is master or the RC branch targetting stable, the job will publish to firebase.

This is not strictly correct. For the stable docs, we should only be publishing after the push to `stable` proper, not the RC branch. To remedy this, I created a new set of two builders in the `//config/packaging_config.star` file, `Linux docs_publish` and `Linux stable docs_publish`. These are the only two builders that are now able to deploy to firebase.

The pre-existing `Linux docs` builders are now renamed `Linux docs_test`, and no longer have the properties necessary to publish to firebase. These are strictly smoke tests, to ensure that the docs correctly compile. This is restoring parity with the setup in Cirrus.


config/framework_config.star
```diff
@@ -351,9 +344,11 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
...
             "validation": "docs",
             "validation_name": "Docs",
-            "dependencies": [{"dependency": "dashing"}, {"dependency": "firebase"}, {"dependency": "curl"}],
-            "firebase_project": firebase_project,
-            "release_ref": release_ref or "refs/heads/master",
+            "dependencies": [{"dependency": "dashing"}, {"dependency": "curl"}],
+            # Test only, the following two keys should be blank, only required
+            # for publishing docs.
+            "firebase_project": "",
+            "release_ref": "",
```
godofredoc, line 351: Would it work if we don't pass them at all?

godofredoc (approved):
LGTM

config/framework_config.star
christopherfujino: Yes, I thought I would explicitly leave them blank with a comment to prevent someone from adding them in the future or with a git merge and unintentionally deploying from these builders.

PR #379, opened 2021-03-06
Remove ios32 tests from staging.

godofredoc:
Removing from staging because we don't have any ios32 devices in
staging.


jmagman (changes requested):
What happened to these devices? @godofredoc signed off on go/flutter-ios32-customers.

jmagman (approved):
Oh STAGING yup LGTM!

PR #380, opened 2021-03-08
Add curl dep to devicelab tasks.

godofredoc:
Bug: https://github.com/flutter/flutter/issues/68242

This is to prevent issues downloading artifacts due to a curl error.


christopherfujino (approved):
LGTM

PR #381, opened 2021-03-09
Mitigate host engine timeouts on presubmit.

godofredoc:
This is to mitigate the timeouts because the windows builds are not
using goma properly.

Bug: https://github.com/flutter/flutter/issues/77665


godofredoc:
Landing as TBR to be able to collect more logs about the failures.

cbracken (commented):
lgtm

PR #383, opened 2021-03-09
Remove all depot_tools/osx_sdk configurations from engine.

godofredoc:
All the flutter repos have been using flutter's xcode bundles for a few
months already. This PR is removing configurations that are not being
used anymore.


keyonghan (approved):
LGTM

PR #384, opened 2021-03-11
update release_branches.json

christopherfujino:
blocked by https://flutter-review.googlesource.com/c/recipes/+/10960

Note that 1.27.x is actually 2.1.x, but the branch was created before the announcement of 2.0. Thus what would have been 1.28.x will actually be 2.2.x.


keyonghan (approved):
LGTM

PR #385, opened 2021-03-11
Separate SDK Linux shards into separate builders

keyonghan:
This create separate builder configs for each subshard (Linux ones).

Related: https://github.com/flutter/flutter/issues/77947


config/framework_config.star
```diff
@@ -675,6 +675,22 @@ def framework_try_config():
     # Defines framework try builders
 
     # Linux platform
+
+    # Configure separate builders for subshards.
+    for subshard in ["1_2", "2_2"]:
```
christopherfujino, line 680: I guess I didn't think this through when I read your design doc, but it's unfortunate that you have to hard-code this into flutter/infra. When this was tracked in the builder properties, it was possible to move that information into the source repo's CI config file.

I suppose one way we can get to the desired state (that is, wasting minimal resources AND allowing the sharding to be configurable in a single PR to the source repo) would be:

1. This change, bringing all the shards to the top level and hard-coding the shard number in flutter/infra
2. @CaseyHillers' change to trigger prod builds from Cocoon, parsing the source repo's CI config file.
3. My change to move the shard number into the properties field of the source repo's config file.

@CaseyHillers & @keyonghan WDYT?

CaseyHillers: SGTM. In the end, we want these types of changes to happen in the source repo so infra doesn't break releases.

FWWIW, my priority for where to put infrastructure logic is:
1. Repo under test
2. flutter/infra
3. flutter/recipes
4. flutter/cocoon

keyonghan: Talked offline with @christopherfujino , this should be the designed state.

godofredoc: +1 to moving the logic to the source repo but that may take some time. We can always deprecate the additional logic as soon as the new functionality becomes available.

config/framework_config.star
```diff
@@ -688,6 +704,22 @@ def framework_try_config():
         caches = LINUX_DEFAULT_CACHES,
         os = LINUX_OS,
     )
+
+    # Configure separate builders for subshards.
+    for subshard in ["libraries", "misc", "widgets"]:
```
godofredoc, line 709: Instead of adding for loops everywhere you may want to create a single function that iterates through the subshard list and creates the builders. E.g.:

def build_with_subshards(shard, subshards, dependencies):
    for shubshard in subshards:
        common.linux_try_builder( ...

if you can make the function very generic you can directly add it to common to be able to use in all the repos if needed.

then everywhere else you can call:
 build_with_subshards("framework_tests" , ["libraries", "misc", "widgets"], deps ....

config/framework_config.star
CaseyHillers: To clarify, I like this change :-) It moves the sharding logic from the recipe to the infra config, which gets us closer to the ideal world of it being controlled in the repo under test.

config/framework_config.star
keyonghan: Used a common function instead: `linux_builder_with_subshards`.

config/framework_config.star
```diff
@@ -675,6 +675,16 @@ def framework_try_config():
     # Defines framework try builders
 
     # Linux platform
+
+    common.linux_builder_with_subshards(
```
godofredoc, line 679: Is it possible to make it even more generic? we'll need the same logic for linux, windows, mac prod and try.

config/lib/common.star
```diff
@@ -393,6 +394,36 @@ _linux_try_builder, _linux_prod_builder = _linux_builder()
...
+def _linux_builder_with_subshards(
+    shard = None,
+    shards = [],
+    dependencies = [],
+    list_view_name = None,
+    caches = None,
+    os = None):
+    for subshard in shards:
+        buildername = "%s_%s" % (shard, subshard)
+        common.linux_try_builder(
```
godofredoc, line 413: What about the prod builders?

godofredoc, line 413: Can we also make the top level function to receive the same parameters, expand the dictionary and only replace the subshard?

config/framework_config.star
```diff
@@ -688,6 +698,15 @@ def framework_try_config():
         caches = LINUX_DEFAULT_CACHES,
         os = LINUX_OS,
     )
+
+    common.linux_builder_with_subshards(
```
godofredoc, line 702: Do we also need to delete the old configs? and what is the strategy to land this change. E.g delete the tasks from try/prod json land this and then add them or the other way aronud?

keyonghan: Here is the workflow: flutter/flutter#77947

config/framework_config.star
keyonghan: Updated. PTAL.

config/lib/common.star
keyonghan: Update. PTAL.

keyonghan: We will implement the prod in separate PRs.

config/framework_config.star
godofredoc: Thanks for the clarification, I thought the comment was about trying to implement all at once.

config/framework_config.star
```diff
@@ -715,6 +762,22 @@ def framework_try_config():
...
+        add_cq = True,
+        list_view_name = list_view_name,
+        properties = {
+            "shard": "tool_integration_tests",
+            "subshards": ["1_3", "2_3", "3_3"],
+            "dependencies": [{"dependency": "android_sdk"}, {"dependency": "chrome_and_driver"}, {"dependency": "open_jdk"}, {"dependency": "goldctl"}, {"dependency": "curl"}],
+        },
+        caches = LINUX_DEFAULT_CACHES,
+        os = LINUX_OS,
+        bucket = "try",
```
godofredoc, line 779: Is this property the only one that is different from prod and try functions?

config/lib/common.star
```diff
@@ -393,6 +393,30 @@ _linux_try_builder, _linux_prod_builder = _linux_builder()
...
 _windows_try_builder, _windows_prod_builder = _windows_builder()
 
+def _short_name(task_name):
+    """Create a short name for task name."""
+    task_name = task_name.replace("__", "_")
+    words = task_name.split("_")
+    return "".join([w[0] for w in words])[:5]
+
+# Configure separate builders for linux subshards.
+def _builder_with_subshards(
```
godofredoc, line 403: docs? this will be used as the entry point on many of the configs and people may want to know what are all the properties about along wit the expected values.

godofredoc (approved):
LGTM with some nits and questions.

config/lib/common.star
keyonghan: Added.

config/framework_config.star
keyonghan: Yes. All others will be passed along as `**kwargs`.

PR #386, opened 2021-03-12
Add prod builders for plugins stable and master.

yusufm:
Bug: 77175

### DIFF
--- a/config/main.star
+++ b/config/main.star
@@ -168,5 +168,5 @@ packaging_config.setup(release_branches)
 
 recipes_config.setup()
 
-plugins_config.setup()
+plugins_config.setup(release_branches)
 ######################### Console Definitions #################################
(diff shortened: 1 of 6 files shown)


config/main.star
```diff
@@ -168,5 +168,5 @@ packaging_config.setup(release_branches)
 
 recipes_config.setup()
 
-plugins_config.setup()
+plugins_config.setup(release_branches)
```
godofredoc, line 171: Should this be a subset? I believe we only want to run for stable and not for dev|beta.

config/plugins_config.star
```diff
@@ -21,6 +21,22 @@ def _setup():
         },
     }
     plugins_define_recipes()
+
+    plugins_prod_config(
```
godofredoc, line 25: if we filter the branches at a higher level then we can use a for loop here.

godofredoc:
Closing as outdated.

PR #387, opened 2021-03-12
Separate SDK Mac try shards into separate builders

keyonghan:
This create separate builder configs for each subshard (Mac try ones).

Related: flutter/flutter#77947


christopherfujino:
Sorry, I think I missed the context. Where are these new builders coming from? I see they have sub-shards defined in the properties...

keyonghan:
> Sorry, I think I missed the context. Where are these new builders coming from? I see they have sub-shards defined in the properties...

They are generated by `common.builder_with_subshards`, which is a general function to support different platforms and pools.

config/framework_config.star
```diff
@@ -977,6 +977,26 @@ def framework_try_config():
...
+            "subshards": ["1_4", "2_4", "3_4", "4_4"],
+            "dependencies": [{"dependency": "android_sdk"}, {"dependency": "chrome_and_driver"}, {"dependency": "open_jdk"}, {"dependency": "xcode"}, {"dependency": "gems"}, {"dependency": "goldctl"}],
+            "$flutter/osx_sdk": {
+                "sdk_version": XCODE_VERSION,
+            },
+        },
+        dimensions = {"device_type": "none"},
+        caches = MAC_NEWXCODE_CACHES,
+        os = MAC_OS,
+        bucket = "try",
```
CaseyHillers, line 998: Why do we specify add_cq=True and bucket="try". Are those equivalent?

keyonghan: `add_cq` is inherited from existing setting (we keep all settings the same as before). It is used to test `recipes` changes (the cq run).
`bucket` is the only extra parameter used to distinguish the pool info when generating configs.

CaseyHillers (approved):
LGTM

PR #389, opened 2021-03-13
Increases timeout of recipe's roller.

godofredoc:
Flutter recipe's have too many dependencies that forces the roller to
check hundredths of revision before selecting the correct one. This has
made the roller to fail consistently over the few last weeks.

Bug: https://github.com/flutter/flutter/issues/78095


godofredoc:
Landing as TBR to unblock the roller.

PR #391, opened 2021-03-16
Separate SDK Linux shards into separate builders - prod

keyonghan:
This create separate builder configs for each subshard (Linux prod ones).

Related: flutter/flutter#77947


config/lib/common.star
```diff
@@ -442,6 +444,13 @@ def _builder_with_subshards(
                 os = os,
                 **kwargs
             )
+        elif bucket == "prod" and os.startswith("Linux"):
```
godofredoc, line 447: if we pass the name can we delete this branch of the code?

keyonghan: Actually the `name` is not needed (skipped from the framework.star side), making it easy here to generate the name based on branch, shards, and subshard. 
(the `name` is also not needed for try builder configs, I plan to remove it in a following PR)

godofredoc: What will be the name if branch_postfix is None? e.g. we are not adding master to the name. Additionally can we call the parameter just branch? it is misleading to use branch_postfix in the middle of the string.

keyonghan: Force inject the parameters. Since `branch` variable already exists, used `branch_name` instead.

godofredoc (approved):

PR #394, opened 2021-03-17
Add device_os dimension to DeviceLab Mac_ios builders

keyonghan:
This adds `device_os` dimension to `Mac_ios` builder configs, so that bigquery buildbucket tables can reveal that info for query analysis. This also makes it consistent across different platforms.

Related issue: https://github.com/flutter/flutter/issues/77389


config/devicelab_staging_config.star
```diff
@@ -162,6 +162,7 @@ def devicelab_staging_prod_config():
             pool = "luci.flutter.staging",
             os = "iOS-14.4",
```
CaseyHillers, line 163: Should we migrate this to macos-$VERSION, and update the scheduler to be based on the os+device_os combo?

CaseyHillers (approved):
LGTM w/ comment on future work

config/devicelab_staging_config.star
keyonghan: good catch. We have been using the combo for linux-android builders, and it would be good to keep them consistent.
Have updated the Mac-android and Mac-ios `os` to reflect host, whereas the `device_os` reflects the device.

config/devicelab_staging_config.star
```diff
@@ -160,7 +163,7 @@ def devicelab_staging_prod_config():
                 "use_cas": True,
             },
             pool = "luci.flutter.staging",
-            os = "iOS-14.4",
+            os = MAC_OS,
             category = "Mac_ios",
```
CaseyHillers, line 167: How do we schedule devicelab ios tasks? Is it based on category? My concern is updating OS will cause there to be no bots to run the tasks.

keyonghan: It is based on the bot `os` and device `device_os` dimension, see values in the bot page, like https://chromium-swarm.appspot.com/bot?id=flutter-devicelab-mac-11.

CaseyHillers: LGTM - Thanks for checking!

PR #396, opened 2021-03-31
Update staging configs to use ios 14.4.2.

godofredoc:
This is in preparation to update all the devicelab phones to that
version.

Bug: https://github.com/flutter/flutter/issues/79431


xster:
LGTM

PR #399, opened 2021-04-01
Update phones to 14.4.2.

godofredoc:
This is to make all the phone versions consistent.

Bug: https://github.com/flutter/flutter/issues/79431


jmagman (approved):
LGTM

PR #400, opened 2021-04-01
Relax the OS dimension to allow M1s and intel machines.

godofredoc:
This only applies to staging environment.


keyonghan (approved):
LGTM

PR #405, opened 2021-04-06
Use bbagent in devicelab tasks.

godofredoc:
This is part of a pending luci migration to the latest runner.

Bug: https://github.com/flutter/flutter/issues/79883


config/devicelab_config.star
```diff
@@ -111,6 +111,7 @@ def devicelab_prod_config(branch, version, ref):
         name = drone_recipe_name,
         cipd_package = "flutter/recipe_bundles/flutter.googlesource.com/recipes",
         cipd_version = "refs/heads/master",
+        use_bbagent = True,
```
CaseyHillers, line 114: Is this expected to be a no-op? Should we split prod and staging into separate PRs?

CaseyHillers (approved):
LGTM

config/devicelab_config.star
godofredoc: Adding the parameter to staging config causes recipe name conflicts .

CaseyHillers: SG thanks for explaining

PR #408, opened 2021-04-07
Add Windows UWP engine shard

cbracken:
Adds a Windows UWP engine build shard. UWP builds are triggered by the
`build_windows_uwp` property.

Related recipe change:
https://flutter-review.googlesource.com/c/recipes/+/12101

Bug: https://github.com/flutter/flutter/issues/78627

PR #409, opened 2021-04-08
Build Windows UWP *only* on the UWP shard

cbracken:
This is a followup to flutter/infra#408, where I declared the parameter,
then utterly failed to actually set it in the properties.

Bug: https://github.com/flutter/flutter/issues/78627


clarkezone (approved):
LGTM

PR #412, opened 2021-04-14
Add another tool_integration_tests subshard

jmagman:
As of https://github.com/flutter/flutter/pull/78649 the longest `tool_integration_tests` subshard went to ~25 minutes on Linux and macOS, and actually timed out on Windows.  Add a fourth subshard.


jmagman:
fyi @jonahwilliams @goderbauer in case you need to change shard number yourselves at some point.
You can have a different number of shard per platform, though in this case I added another `tool_integration_tests` subshard for all platforms so they are still the same.

jonahwilliams (approved):
It would be great if you could update the flutter_tools README with the instructions on how to do this (and maybe a link to this PR)

LGTM

jmagman:
I just remembered this isn't going to work without also updating the [try and prod builders](https://github.com/flutter/flutter/blob/5140376ee34fc030ba1c5f1f1e637ed6c702547d/dev/try_builders.json#L199) in the framework, which means that the release shards will break because the old ones don't exist.

Hmmm... Really hard to do this when it's not versioned between channels and it's split between repos.

jmagman:
@keyonghan what do you recommend here now that the shard names are directly in the framework as of https://github.com/flutter/flutter/issues/77947?  Before, the `"Linux tool_integration_tests` would just collect all of them and the framework didn't need to know how many subshards there are.

jmagman:
> @keyonghan what do you recommend here now that the shard names are directly in the framework as of [flutter/flutter#77947](https://github.com/flutter/flutter/issues/77947)? 

Maybe remove the old builders from the json files in the framework, rename the builders in infra with this PR, then add the new names back to the framework once the new builders exist?  And cherry pick those name changes into the release branches?

keyonghan:
> @keyonghan what do you recommend here now that the shard names are directly in the framework as of [flutter/flutter#77947](https://github.com/flutter/flutter/issues/77947)? Before, the `"Linux tool_integration_tests` would just collect all of them and the framework didn't need to know how many subshards there are.

We can do the following
1) Keep existing subshard setting (1_3, 2_3, 3_3), so that it will not break anything
2) Add new set of subshards (1_4, ..., 4_4) here in config
3) Update prod_builders.json/try_builders.json with 4 shards, and remove 3 shards
4) remove 1_3, ..., 3_3 configs from here

jmagman:
@keyonghan The entire point of that `1_4` naming mechanism was to keep the framework ignorant of the number of subshards that exist, so it could just be changed in the config without anything needing to change in the tests.  These cross-repo atomic changes are excruciating. 

Can we change the concept of subshards from the title back to be an env `SUBSHARD` that gets set instead?  And the name of the task that needs to change in the frameworks contains just the numerator? `tool_integration_tests_1` `tool_integration_tests_2` etc?

https://github.com/flutter/flutter/blob/c571804913c189479ba30bdebceaececcafcd512/dev/bots/test.dart#L1451-L1456
```
/// For local testing you can just set the SHARD and SUBSHARD
/// environment variables. For example, to run all the framework tests you can
/// just set SHARD=framework_tests. Some shards support named subshards, like
/// SHARD=framework_tests SUBSHARD=widgets. Others support arbitrary numbered
/// subsharding, like SHARD=build_tests SUBSHARD=1_2 (where 1_2 means "one of two"
/// as in run the first half of the tests).
```

keyonghan:
We may encounter issues if we update the builder name to be independent of the shard number, namely only based on numerator.

When we increase shards from 3 to 4:
Before: test_1 (mapped to subshard 1_3), test_2 (mapped to subshard 2_3), test_3 (mapped to subshard 3_3)
After: test_1 (mapped to subshard 1_4), test_2 (mapped to subshard 2_4), test_3 (mapped to subshard 3_4), test_4 (mapped to subshard 4_4)
In this case, if we change the config side first from 3 to 4, the test_4 will not be executed before we land the change in .json file. So we will have a gap in between which misses the subshard 4_4.

When we decrease shards from 4 to 3:
Before: test_1 (mapped to subshard 1_4), test_2 (mapped to subshard 2_4), test_3 (mapped to subshard 3_4), test_4 (mapped to subshard 4_4)
After: test_1 (mapped to subshard 1_3), test_2 (mapped to subshard 2_3), test_3 (mapped to subshard 3_3)
In this case, if we change the config side first from 4 to 3, the test_4 will cause problem because the config for test_4 has been removed from config side.
If we choose to make changes from .json file first from 4 to 3, then we will face the same issue as the first case: missing the subshard 4_4.

At this point I don't see an easy way to keep framework ignorant of number of subshard but maintain separated subshard builders.

Like said earlier, one possibility is to keep existing configs (either from 4 to 3, or from 3 to 4), and add the new subshards config. Once ready, update the json files and clean up the old config..  This way, we will not loss any test, but will run duplicated test on the same `SHARD` during the transition.

jmagman:
> Like said earlier, one possibility is to keep existing configs (either from 4 to 3, or from 3 to 4), and add the new subshards config.

Thanks for the analysis @keyonghan.  Updated to include both the old and new names.

keyonghan (approved):
LGTM

PR #413, opened 2021-04-14
Revert back to using luciexe instead of bbagent.

godofredoc:
This is to try to find out why flutter doctor and in general any steps
using pub get are timing out very frequently.

Bug: https://github.com/flutter/flutter/issues/80317


jmagman (approved):
RSLGTM

PR #415, opened 2021-04-15
Remove tool_integration_tests subshards

jmagman:
Turns out keeping the old *_3 subshards wasn't necessary for release branches from #412.


christopherfujino (approved):
Thanks!

PR #422, opened 2021-04-23
Add skp_generator shard to flutter/flutter

Hixie:
Fixes https://github.com/flutter/flutter/issues/80906
Depends on https://github.com/flutter/flutter/pull/81087


Hixie:
cc @christopherfujino

config/framework_config.star
```diff
@@ -473,6 +473,23 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
         caches = LINUX_DEFAULT_CACHES,
         os = LINUX_OS,
     )
+    if branch == "master":
```
christopherfujino, line 476: nice

config/framework_config.star
```diff
@@ -473,6 +473,23 @@ def framework_prod_config(branch, version, testing_ref, release_ref):
...
+        common.linux_prod_builder(
+            name = "Linux%s skp_generator|skp_gen" % ("" if branch == "master" else " " + branch),
+            recipe = drone_recipe_name,
+            console_view_name = console_view_name,
+            triggered_by = [trigger_name],
+            triggering_policy = triggering_policy,
+            priority = priority,
+            properties = {
+                "shard": "skp_generator",
+                "subshard": "0",
```
christopherfujino, line 486: @keyonghan is this the right thing to do?

christopherfujino: or does it not matter since this will be parsed by test.dart?

Hixie: FWIW, there's no subshards in test.dart for this shard, so whatever we pass in as SUBSHARD will do the same thing. Happy to put something else in here though.

keyonghan: This is okay since we have only one subshard (similar case as [flutter_plugins](https://devops-console-oss.corp.google.com/flutter/infra/+/master:config/framework_config.star;l=460)). 
When we have multiple subshards, `common. builder_with_subshards` will be the way to go.

christopherfujino: yeah makes sense, as long test.dart ignores this field, it doesn't matter what we put here.

christopherfujino (approved):
lgtm

PR #425, opened 2021-04-28
Add another tool_integration_tests subshard for Windows

goderbauer:
The longest Windows `tool_integration_tests` shard is timing out:
https://ci.chromium.org/ui/p/flutter/builders/prod/Windows%20tool_integration_tests_1_4/172/overview
https://ci.chromium.org/ui/p/flutter/builders/prod/Windows%20tool_integration_tests_1_4/173/overview

Adding another shard...


goderbauer:
Actually, does it work to only add the shard to Windows or should Mac and Linux also get another one (they don't appear to be close to the timeout yet)?

config/framework_config.star
```diff
@@ -1216,7 +1217,8 @@ def framework_try_config():
         list_view_name = list_view_name,
         properties = {
             "shard": "tool_integration_tests",
-            "subshards": ["1_4", "2_4", "3_4", "4_4"],
+            # TODO(goderbauer): Remove *_4 subshards once the framework adopts *_5 builders in stable.
```
jmagman, line 1220: It turns out we only need to wait until https://github.com/flutter/flutter/pull/81400 lands to remove the old builders.  The framework `_builders.json` files are only used for the build dashboard, but the release branches don't use that, they use the milo dashboard.  So the new builder will work on release branches even with the json numbering mismatch.

jmagman (approved):
LGTM! You can leave the TODO as-is if you want to merge this ASAP to get the build green, but we can remove the old builder numbers once https://github.com/flutter/flutter/pull/81400 merges.

PR #430, opened 2021-05-03
Use only 12 core macs in tool integration tests.

godofredoc:
Using 4 core vms have been causing those tests to fail consistently.

Bug: https://github.com/flutter/flutter/issues/81793


Piinks (approved):
LGTM

keyonghan:
Shall we make it the same for try builders?

godofredoc:
Let me check if the try pool has 4 core machines

godofredoc:
> Let me check if the try pool has 4 core machines

Actually not needed we only have 12 core machines in the try pool.

PR #431, opened 2021-05-04
Add cores dimension to the correct place.

godofredoc:
The previous PR added the dimension to the linux bots rather than the
mac ones.

Bug: https://github.com/flutter/flutter/issues/81793


godofredoc:
TBR to reopen the tree

PR #432, opened 2021-05-04
Turned on the platform channels benchmarks

gaaclarke:
Don't land until https://github.com/flutter/flutter/pull/81414 lands.


config/devicelab_config.star
```diff
@@ -464,6 +465,7 @@ def devicelab_prod_config(branch, version, ref):
         "new_gallery__transition_perf",
         "picture_cache_perf__timeline_summary",
         "platform_channel_sample_test",
+        "platform_channels_benchmarks",
```
keyonghan, line 468: Left some comments in https://github.com/flutter/flutter/pull/81414.

Considering the above PR, shall this one be added under mac_ios? Or the other way around change the name in the above PR to be mac_android?

gaaclarke: I addressed the naming comments in https://github.com/flutter/flutter/pull/81874 since it was about to land.

Also moved it to mac_ios, nice catch, thanks!

keyonghan (approved):
LGTM

gaaclarke:
Oh shoot, should i have waited for the renaming to land before I landed this?

keyonghan:
> Oh shoot, should i have waited for the renaming to land before I landed this?

This is okay.
The changes here only make the builders visible to LUCI world when the builders are triggered.

https://github.com/flutter/flutter/pull/81414 has added corresponding tests, and this will make the builder config here know what to run.

The pending PR: https://github.com/flutter/flutter/pull/81874 only renames the builders, and it affects only tests presentation on flutter build dashboard. So after this PR lands, you will expect builder info to be refreshed in the dashboard.

PR #436, opened 2021-05-06
Remove the certs.

godofredoc:
The certs step didn't work because it required a change in the dart sdk
that is still on review.

Bug: https://github.com/flutter/flutter/issues/81837


godofredoc:
Landing as TBR

zanderso (commented):
lgtm

PR #440, opened 2021-05-07
Temporarily disable UWP builders to re-open tree

cbracken:
Due to LUCI bot image updates, our build bots no longer include a recent
Windows SDK, which prevents the UWP embedder from building.

There is separate work in progress as part of flutter/flutter#81994 to
get an updated SDK on all Windows bots, at which point we will reinstate
the UWP build.

See: https://github.com/flutter/flutter/issues/81994


cbracken:
Closing in favour of @aam's update to his patch here: https://github.com/flutter/engine/pull/25993

PR #442, opened 2021-05-07
Enable Windows Host Engine as a recipes presubmit.

godofredoc:
Recipe changes are not being validated on windows bots. This change
will run Windows Host Engine builds on recipe changes.


cbracken (approved):
LGTM stamp from a Japanese personal seal

PR #444, opened 2021-05-11
remove web e2e and smoke test builders

yjbanov:
They have been rolled under the web_long_running_tests shard.


keyonghan (approved):
LGTM

PR #445, opened 2021-05-12
Consolidate web integration tests

yjbanov:
These tests are merged into the `web_long_running_tests` shard.


keyonghan (commented):
To avoid break, we need to add subshards config ["1_5", ..., "5_5"] for `Linux%s web_long_running_tests` both `try` and `prod` first. Then land https://github.com/flutter/flutter/pull/82307/files.
After validated, then remove the subshards config ["1_3", "2_3", "3_3"] for existing `Linux%s web_long_running_tests`.

yjbanov:
Good catch! Done.

keyonghan (approved):
LGTM

PR #452, opened 2021-05-13
Fix dependency key.

godofredoc:
Windows tool_integration_tests dependency key has several white spaces
at the end of the string causing the dependency installation to fail.


godofredoc:
Landing as TBR to fix failures installing the dependencies.

PR #453, opened 2021-05-13
Add prod builders for plugins repo.

godofredoc:
Bug: https://github.com/flutter/flutter/issues/77175


config/plugins_config.star
```diff
@@ -73,6 +80,75 @@ def plugins_try_config(platform_args):
         **platform_args["windows"]
     )
 
+######### plugins prod config
+def plugins_prod_config(branch, version, testing_ref, release_ref, platform_args):
+    """Prod configurations for the plugins repository.
+
+    Args:
+      branch(str): The branch name we are creating configurations for.
+      version(str): One of dev|beta|stable.
```
stuartmorgan, line 89: The two configurations we usually test for plugins are master and stable; should master be listed here?

config/plugins_config.star
```diff
@@ -73,6 +80,75 @@ def plugins_try_config(platform_args):
...
+        name = trigger_name,
+        bucket = "prod",
+        repo = repos.PLUGINS,
+        refs = [testing_ref],
+    )
+
+    # Defines triggering policy
+    if branch == "master":
+        triggering_policy = scheduler.greedy_batching(
+            max_batch_size = 3,
```
stuartmorgan, line 128: Does this mean multiple PRs will be tested at the same time?

config/generated/flutter/luci/luci-milo.cfg
```diff
@@ -6310,4 +6310,17 @@ consoles {
   }
   favicon_url: "https://storage.googleapis.com/flutter_infra/favicon.ico"
 }
+consoles {
+  id: "plugins"
+  name: "plugins"
+  repo_url: "https://chromium.googlesource.com/external/github.com/flutter/flutter"
```
stuartmorgan, line 6316: Does the fact that this says flutter/flutter rather than flutter/plugins mean there's a problem with the definition?

stuartmorgan (commented):
I have almost no experience with these files, so I suspect my review isn't very useful. Apologies if the questions are all just from me not understanding the files.

config/plugins_config.star
godofredoc: yes, I copy pasted this from the framework but we can make it 1 for plugins.

config/plugins_config.star
godofredoc: Updated the doc, we are currently adding master, how do you define stable for plugins? e.g. this is the flutter/flutter definition https://cs.opensource.google/flutter/infra/+/master:config/lib/release_branches/branches.json;l=2

config/generated/flutter/luci/luci-milo.cfg
godofredoc: nice catch, checking where is this coming from.

godofredoc: updated, I was using flutter repo instead of plugins on the luci view definition.

config/plugins_config.star
stuartmorgan: We run `flutter channel stable` followed by `flutter upgrade`. I.e., it's always the latest version of stable.

We could revisit that in theory, but not testing what we expect most users to be running the plugins against would not be ideal.

config/plugins_config.star
stuartmorgan: Windows changes are rare enough it's probably okay with 3, but I'd rather we start with 1 and see if it's an issue.

config/plugins_config.star
godofredoc: I finally understand the request, basically we want to run two builders one for flutter/flutter master and one for flutter/flutter stable for every commit in plugins master. Updating the recipes to pass a property for this.

PR #454, opened 2021-05-14
Add build_integration tests to led presubmit tests.

godofredoc:
This is to be able to validate the vs installation process.


keyonghan (approved):
LGTM

PR #457, opened 2021-05-18
Update plugins recipes with the plugins prefix.

godofredoc:
The plugins recipes are inside a plugins folder that was missing in the
build definitions.


config/plugins_config.star
```diff
@@ -97,13 +97,14 @@ def plugins_prod_config(branch, version, testing_ref, release_ref, platform_args
...
     """
-    recipe_name = ("plugins_" + version if version else "plugins")
-    luci.recipe(
-        name = recipe_name,
-        cipd_package = "flutter/recipe_bundles/flutter.googlesource.com/recipes",
-        cipd_version = "refs/heads/master",
-        use_bbagent = True,
-    )
+    recipe_name = ("plugins/plugins_" + version if version else "plugins/plugins")
+    if branch != "master":
```
yusufm, line 101: Does branch = master need to be handled somewhere?

yusufm (approved):
LGTM with one comment

config/plugins_config.star
keyonghan, line 101: Is this for non master branch?
(https://ci.chromium.org/p/flutter/g/plugins/console is for master).

godofredoc: It is handled in plugins_define_recipes above.

godofredoc: No this is only for master. Plugins only supports master.

PR #462, opened 2021-05-25
Add VS and Certs dependencies to plugins.

godofredoc:
This is to prevent TLS issues with missing certificates.

Bug: https://github.com/flutter/flutter/issues/83349


CaseyHillers (approved):
Should this be added to all windows builders?

stuartmorgan (approved):
LGTM.

Do we understand how this regressed without anything flagging it until PRs started failing?

godofredoc:
> nd how this regressed without anything flagging it until PRs started failing?

Yes, we actually never added the certs dependency to plugins. We were being lucky that by the time plugins commits were running tests the windows bots were already used by framework|engine tasks and the certs were available.

PR #464, opened 2021-05-26
Add dependencies for plugins try builder.

godofredoc:
The dependencies were added to the prod builders but we missed the try
builder.


stuartmorgan (approved):
LGTM

PR #467, opened 2021-06-04
more time for devicelab tasks

dnfield:
We currently let the main body of these tests take 30 minutes. If that times out, all subsequent steps fail to run becaue of global shutdown. This means we don't get a chance to upload logs or screenshots.

see https://github.com/flutter/flutter/issues/83939


Hixie (approved):
RSLGTM

Hixie:
test-exempt: configuration

PR #468, opened 2021-06-04
Add property for whether a device is attached

dnfield:
To enable https://flutter-review.googlesource.com/c/recipes/+/14400


config/devicelab_config.star
```diff
@@ -392,6 +392,7 @@ def devicelab_prod_config(branch, version, ref):
                 "task_name": task,
                 "upload_metrics": branch in UPLOAD_METRICS_CHANNELS,
                 "use_cas": True,
+                "attached_device": True,
```
godofredoc, line 395: Can we use device_os instead? if device_os is none then it does not have any device attached.

dnfield: We can't access that from the recipe without making a network request to swarming.

dnfield: (Tagged you in a relevant LUCI-users thread)

godofredoc: Thanks!

godofredoc (approved):

PR #469, opened 2021-06-07
Cache OSX_SDK on mac builders

iskakaushik:


godofredoc (approved):

iskakaushik:
cc: @zanderso

zanderso:
Cool! This appears to be working sometimes:

https://ci.chromium.org/ui/p/flutter/builders/prod/Mac%20Web%20Engine/5403/overview

I guess it will take some time for the cache to be warmed up everywhere. For example not yet warm here:

https://ci.chromium.org/ui/p/flutter/builders/prod/Mac%20Web%20Engine/5407/overview

godofredoc:
We probably need one named cache for the engine and a different one for framework, if we use the same name for both they will be overriding each other with versions 12.4(framework) and 11.2(engine).

PR #470, opened 2021-06-09
Use old_osx_sdk name for engine xcode caches.

godofredoc:
This is to avoid the caches to be overwritten when bots run tasks
alternating from framework and engine.


iskakaushik (approved):
LGTM

PR #471, opened 2021-06-15
Override builder cache to remove the 4m latency to get a bot.

godofredoc:
Buildbucket adds a 4m latency by default that adds 4 mins on top of the
execution time. In the flutter use case we would like to pick up the
next available bot if there is no one with the requested caches
immediately.

Bug: https://github.com/flutter/flutter/issues/84295


iskakaushik (approved):
LGTM

PR #476, opened 2021-06-23
Generate cocoon builders from ci.yaml

CaseyHillers:
- Add `config/cocoon_config.json` (JSON version of `.ci.yaml` in flutter/cocoon)
- Add `ci_yaml` module to parse the json and generate builders from it

# Issues
https://github.com/flutter/flutter/issues/84998

# Future work
Create a builder that updates `scheduler.proto` and `cocoon_config.json`


config/cocoon_config.json
```diff
@@ -0,0 +1 @@
+{"targets":[{"name":"Linux Cocoon","builder":"Linux Cocoon","scheduler":"luci","postsubmit":false,"recipe":"cocoon/cocoon"},{"name":"Linux device_doctor","builder":"Linux device_doctor","scheduler":"luci","postsubmit":false,"runIf":["device_doctor/**"],"recipe":"cocoon/device_doctor"},{"name":"Mac device_doctor","builder":"Mac device_doctor","scheduler":"luci","postsubmit":false,"runIf":["device_doctor/**"],"recipe":"cocoon/device_doctor"},{"name":"Windows device_doctor","builder":"Windows device_doctor","scheduler":"luci","postsubmit":false,"runIf":["device_doctor/**"],"recipe":"cocoon/device_doctor"}],"enabledBranches":["master"]}
```
CaseyHillers, line 1: Future work includes pretty printing these files so it's easier to read by humans

config/generated/flutter/luci/cr-buildbucket.cfg
```diff
@@ -51451,7 +51451,6 @@ buckets {
         properties_j: "$recipe_engine/isolated:{\"server\":\"https://isolateserver.appspot.com\"}"
         properties_j: "$recipe_engine/swarming:{\"server\":\"https://chromium-swarm.appspot.com\"}"
         properties_j: "clobber:false"
-        properties_j: "fuchsia_ctl_version:\"version:0.0.23\""
```
CaseyHillers, line 51454: This shouldn't be relevant for Cocoon, but I suspect this is surfacing in issue in how the linux builders are being generated

config/lib/ci_yaml/ci_yaml.star
```diff
@@ -0,0 +1,87 @@
...
+        target_args = {
+            "add_cq": True,
+            "execution_timeout": target.timeout * time.minute,
+            "list_view_name": list_view_name,
+            "name": "%s|%s" % (target.name, common.short_name(target.name)),
+            "recipe": target.recipe,
+            "repo": repos.SOURCE[repo_name],
+            "properties": target.properties,
+        }
+        if target.name.startswith("Linux"):
```
CaseyHillers, line 63: This block seems like a lot of duplication. If anyone familiar with starlark has recommendations please let me know!

config/generated/flutter/luci/cr-buildbucket.cfg
christopherfujino: nice

config/cocoon_config.star
```diff
@@ -9,9 +9,7 @@ The schedulers pull commits indirectly from GoB repo (https://chromium.googlesou
 which is mirrored from https://github.com/flutter/cocoon.
 """
 
-load("//lib/common.star", "common")
-load("//lib/repos.star", "repos")
-load("//lib/timeout.star", "timeout")
+load("//lib/ci_yaml/ci_yaml.star", "ci_yaml")
 
 def _setup():
     platform_args = {
```
christopherfujino, line 15: Doesn't have to be in this commit, but is there any reason not to move this into .ci.yaml?

config/cocoon_config.star
```diff
@@ -22,73 +20,7 @@ def _setup():
...
-    common.windows_try_builder(
-        name = "Windows device_doctor|device_doctor",
-        recipe = "cocoon/device_doctor",
-        list_view_name = list_view_name,
-        repo = repos.COCOON,
-        add_cq = True,
-        execution_timeout = timeout.SHORT,
-        **platform_args["windows"]
-    )
+    ci_yaml.recipes("cocoon")
```
godofredoc, line 23: Is it possible to override the recipe in the buildbucket calls?

config/cocoon_config.star
```diff
@@ -22,73 +20,7 @@ def _setup():
         "windows": {"os": "Windows"},
         "mac": {"os": "Mac-10.15"},
     }
```
godofredoc, line 22: rm if we will pass this to the buildbucket call?

config/lib/ci_yaml/ci_yaml.star
godofredoc: We can potentially remove most of this code. We only need a basic template that we can call in a loop with different names.

For try we only need a single builder that will be overwritten with the call to buildbucket. Prod builds will only need the name with the most basic config as these files will be used only to get the name of the builder when populating milo dashboards.

config/lib/ci_yaml/compile_proto.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
christopherfujino, line 1: Can you add a copyright comment on this (and https://github.com/flutter/infra/blob/master/config/lib/release_branches/compile_proto.sh which I also didn't put a copyright on)?

christopherfujino (commented):
Did you try this on a bigger .ci.yaml file, such as the Framework? Not that we need to commit it now, but to verify ci_yaml.star supports everything we need?

CaseyHillers:
> Did you try this on a bigger .ci.yaml file, such as the Framework? Not that we need to commit it now, but to verify ci_yaml.star supports everything we need?

Yes. Not everything is supported (I need to make target.name == target.builderName, port properties into .ci.yaml, ensure builder name contains the platform info)

config/cocoon_config.star
CaseyHillers: Is it okay to punt this to a future PR? It makes sense to add these to .ci.yaml and pass that here (once available)

config/cocoon_config.star
CaseyHillers: Yes, but I believe they need to exist in this config to be available (we can only specify a recipe once)

config/lib/ci_yaml/compile_proto.sh
CaseyHillers: Done

CaseyHillers:
> We can potentially remove most of this code. We only need a basic template that we can call in a loop with different names.

Done! I use the common builder to simplify the logic.

config/cocoon_config.star
godofredoc: SGTM

config/cocoon_config.json
godofredoc: Ideally humans won't need to check these files very frequently.

config/cocoon_config.star
godofredoc: Yeah, you are right. The recipe should be registered before it can be used in a builder.

config/generated/flutter/luci/cr-buildbucket.cfg
godofredoc: Yeah, this is just a default being assigned somewhere in the common.star file.

config/lib/repos.star
```diff
@@ -25,4 +25,7 @@ repos = struct(
     OPENSSL = "https://flutter-mirrors.googlesource.com/openssl",
     IOS_DEPLOY = "https://flutter-mirrors.googlesource.com/ios-deploy",
     LIBZIP = "https://flutter-mirrors.googlesource.com/libzip",
+    SOURCE = {
```
godofredoc, line 28: Can we use a more descriptive name and/or add a comment here?

config/lib/ci_yaml/compile_proto.sh
```diff
@@ -0,0 +1,17 @@
...
+# Copyright 2020 The Flutter Authors. All rights reserved.
+# Use of this source code is governed by a BSD-style license that can be
+# found in the LICENSE file.
```
godofredoc, line 4: Do we have a readme explaining how to use this file?

godofredoc (approved):
Some questions and nits, looks good to me after updating comments and answering the questions.

config/lib/repos.star
CaseyHillers: Switched to GIT_REMOTE

config/lib/ci_yaml/compile_proto.sh
CaseyHillers: Added!

PR #477, opened 2021-06-23
[hhh] Add devicelab try builder

CaseyHillers:
This will allow the Dart team to experiment with hot_mode_dev_cycle_linux__benchmark on Android. They can verify their pipelines on their side before they use the existing prod builders.


CaseyHillers:
\cc @whesse

config/devicelab_staging_config.star
```diff
@@ -325,4 +325,46 @@ def devicelab_staging_prod_config():
...
+                },
+            ],
+            "task_name": "hot_mode_dev_cycle_linux__benchmark",
+            "upload_metrics": False,
+            "use_cas": True,
+            "attached_device": True,
+        },
+        os = "Android",
+        dimensions = {"device_os": "N"},
+        # TODO(keyonghan): adjust the timeout when devicelab linux tasks are stable:
```
godofredoc, line 360: Seems like this TODO is outdated

godofredoc: We should probably remove it, the associated bug is closed.

config/lib/repos.star
```diff
@@ -16,6 +16,8 @@ repos = struct(
         "https://chromium.googlesource.com/external/github.com/flutter/engine",
     PACKAGES =
         "https://chromium.googlesource.com/external/github.com/flutter/packages",
+    DART_HHH =
```
godofredoc, line 19: How are we planning to trigger the build?

CaseyHillers: Dart will be triggering these builds when they run their HHH builds through buildbucket. This repo logic is part of the try_builder boiler plate, and will not be used by these try builds.

config/devicelab_staging_config.star
```diff
@@ -357,13 +357,9 @@ def devicelab_staging_prod_config():
         },
         os = "Android",
         dimensions = {"device_os": "N"},
-        # TODO(keyonghan): adjust the timeout when devicelab linux tasks are stable:
-        # https://github.com/flutter/flutter/issues/72383.
-        expiration_timeout = timeout.XL,
         execution_timeout = timeout.MEDIUM,
         caches = LINUX_DEFAULT_CACHES,
         category = "Linux_android",
-        pool = "luci.flutter.try",
```
godofredoc: pool = 'luci.flutter.staging'?

godofredoc (approved):
LGTM after adding back the pool property and updating it to point to staging.

